PR #14, opened 2019-09-19
feat: allow custom login and logout paths

joshcanhelp:
### Description

Allows custom login and logout paths in the same way as custom callback ones. See API.md for more information. Custom routes can also be done [this way](https://github.com/auth0/express-openid-connect/blob/master/EXAMPLES.md#example-2).

### Testing

### DIFF
--- a/API.md
+++ b/API.md
@@ -14,12 +14,15 @@ In general, you won't need to configure this middleware besides the required par
 | clientSecret        | `env.CLIENT_SECRET`             | The client secret, only required for some grants.                              |
 | clockTolerance      | `5`                             | The clock's tolerance in seconds for token verification.                       |
 | getUser             | `tokenSet => tokenSet.claims()` | An async function receiving a tokenset and returning the profile for `req.openid.user`. |
-| required            | `true`                            | If true requires authentication for all the routes in the stack. You can also provide a function to determine if is required based on the request.               |
+| required            | `true`                          | If true requires authentication for all the routes in the stack. You can also provide a function to determine if is required based on the request.               |
 | handleUnauthorizedErrors | `true`                     | Install a middleware that handles Unauthorized/401 errors by triggering the login process. |
 | routes              | `true`                          | Installs the `GET /login` and `GET /logout` route.                              |
 | idpLogout           | `false`                         | Logout the user from the identity provider on logout                            |
 | auth0Logout         | `false`                         | Enable Auth0's non-compliant logout feature, only if Auth0 can be detected and the Auth0 instance does not support OpenID Connect session management. |
-| authorizationParams | See bellow                      | The parameters for the authorization call.                                      |
+| redirectUriPath     | `/callback`                     | Relative path for the callback URL; sent to the authroize endpoint in the `redirectUri` parameter. |
+| loginPath           | `/login`                        | Relative path to login.                                                         |
+| logoutPath          | `/logout`                       | Relative path to logout.                                                        |
+| authorizationParams | See below                       | The parameters for the authorization call.                                      |
 
 Default value for `authorizationParams` is:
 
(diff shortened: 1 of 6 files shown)


API.md
```diff
@@ -27,12 +27,15 @@ In general, you won't need to configure this middleware besides the required par
...
+| required            | `true`                          | If true requires authentication for all the routes in the stack. You can also provide a function to determine if is required based on the request.               |
 | handleUnauthorizedErrors | `true`                     | Install a middleware that handles Unauthorized/401 errors by triggering the login process. |
 | routes              | `true`                          | Installs the `GET /login` and `GET /logout` route.                              |
 | idpLogout           | `false`                         | Logout the user from the identity provider on logout                            |
 | auth0Logout         | `false`                         | Enable Auth0's non-compliant logout feature, only if Auth0 can be detected and the Auth0 instance does not support OpenID Connect session management. |
-| authorizationParams | See bellow                      | The parameters for the authorization call.                                      |
+| redirectUriPath     | `/callback`                     | Relative path for the callback URL; sent to the authroize endpoint in the `redirectUri` parameter. |
```
lbalmaceda:
```suggestion
| redirectUriPath     | `/callback`                     | Relative path for the callback URL; sent to the authorize endpoint in the `redirectUri` parameter. |
```

API.md
```diff
@@ -27,12 +27,15 @@ In general, you won't need to configure this middleware besides the required par
 | clientSecret        | `env.CLIENT_SECRET`             | The client secret, only required for some grants.                              |
 | clockTolerance      | `5`                             | The clock's tolerance in seconds for token verification.                       |
 | getUser             | `tokenSet => tokenSet.claims()` | An async function receiving a tokenset and returning the profile for `req.openid.user`. |
-| required            | `true`                            | If true requires authentication for all the routes in the stack. You can also provide a function to determine if is required based on the request.               |
+| required            | `true`                          | If true requires authentication for all the routes in the stack. You can also provide a function to determine if is required based on the request.               |
 | handleUnauthorizedErrors | `true`                     | Install a middleware that handles Unauthorized/401 errors by triggering the login process. |
 | routes              | `true`                          | Installs the `GET /login` and `GET /logout` route.                              |
```
lbalmaceda: wouldn't this change with this PR as well?

lib/config.js
```diff
@@ -29,6 +29,8 @@ const paramsSchema = Joi.object().keys({
   errorOnRequiredAuth: Joi.boolean().optional().default(false),
   auth0Logout: Joi.boolean().optional().default(false),
   redirectUriPath: Joi.string().optional().default('/callback'),
+  loginPath: Joi.string().optional().default('/login'),
```
lbalmaceda: I think you're assuming the `/` is always included in the value, shouldn't you validate that?

joshcanhelp: Probably not a bad idea. I'm using the same convention as the callback but it's probably good to do for all 3. Good catch.

#21

API.md
joshcanhelp: No. This switch makes authentication required on all routes when it's `true` (default but considering whether that's best) or not (using `requireAuth()`) when it's `false`.

lbalmaceda: I mean the description of the property. If the /login and /logout paths can now be customized, maybe it's worth pointing out those new properties refer to that value?

middleware/auth.js
```diff
@@ -68,10 +68,10 @@ module.exports = function (params) {
   });
 
   if (config.routes) {
-    router.get('/login', (req, res) => {
+    router.get(config.loginPath, (req, res) => {
```
lbalmaceda: I think it's missing the tests for this file changes. You are now testing the config options are being set, but you're not testing that they are being used in the router.

joshcanhelp:
@lbalmaceda - Ready to re-review when you're able. Thank you!

PR #15, opened 2019-09-20
feat: add support for other ID token algorithms

joshcanhelp:
### Description

Add support for HS256 applications and ensures that the ID token algorithm is supported by the Issuer.

### Testing


joshcanhelp:
Thanks @stevehobbsdev!

Just a quick note: we're still discussing internally whether this should be a part of the package or not. Opinions welcome!

joshcanhelp:
@stevehobbsdev - Rebase dismissed your review.

PR #16, opened 2019-09-24
Update README, EXAMPLE, and API documentation

joshcanhelp:


API.md
```diff
@@ -53,24 +40,25 @@ Commonly used `authorizationParams`:
...
-## openidClient.requiresAuth options
+## openidClient.requiresAuth
 
-The middleware doesn't have any options
+The `requiresAuth()` middleware protects specific application routes:
 
 ```javascript
 const { auth, requiresAuth } = require('express-openid-connect');
-app.use('/admin', requiresAuth());
+app.use( auth( { required: false } ) );
```
albertoperdomo: From what I understand, `required` is `true` by default. Does it make sense to use non-default in this example to make things simpler?

I would rather like to see the simplest example to begin with, then go on to special cases after.

Nothing major that should hold us back at this time. Feel free to ignore.

joshcanhelp: I'm considering whether `required` should be on our not ... seems like it should be off by default. Either way, i'll make sure to adjust these appropriately. 

This doc is more about whats' possible. The EXAMPLES on in this changeset is more about how to get started.

PR #17, opened 2019-09-24
Remove debugging callbacks

joshcanhelp:
Waiting on #12


joshcanhelp:
@damieng - Agreed but this came with an `npm uninstall` so the lock file was altered.

PR #27, opened 2019-10-08
add typescript types

jbarrus:

### Description

Adds typescript definitions to allow easier use with typescript projects.

### References

https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html

### Testing

Testing can be done by importing into a typescript project.

### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,22 +3,25 @@
 import { AuthorizationParameters, TokenSet } from 'openid-client';
 import { Request, RequestHandler } from 'express';
 
-interface AuthParams {
-    issuerBaseURL?: string;
+interface ConfigParams {
+    auth0Logout?: boolean;
+    authorizationParams?: AuthorizationParameters
     baseURL?: string;
     clientID?: string;
     clientSecret?: string;
     clockTolerance?: number;
-    getUser?: (tokenSet: TokenSet) => any;
-    required?: boolean | ((request: Request) => boolean);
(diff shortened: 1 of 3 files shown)


joshcanhelp:
@jbarrus - Thank you so much for this! We'll take a look this week and get back to you with feedback or an approval.

index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+import { AuthorizationParameters, TokenSet } from 'openid-client';
+import { Request, RequestHandler } from 'express';
+
+interface AuthParams {
+    issuerBaseURL?: string;
+    baseURL?: string;
+    clientID?: string;
+    clientSecret?: string;
+    clockTolerance?: number;
+    getUser?: (tokenSet: TokenSet) => any;
```
joshcanhelp: Should this really be `any`? This function should return `undefined` or an `object` representing the user. Can we be more explicit here?

jbarrus: Does it have to be an object or can it be a string or something else?

joshcanhelp: Here's the default `getUser` implementation:

https://github.com/auth0/express-openid-connect/blob/master/lib/getUser.js

We should definitely be strict on what it gets back. Following this up to the parent library, by default, that returns `IdTokenClaims`, which extends `UserinfoResponse`. I would probably use the latter for now.

Issue #35, opened 2019-11-19
How to provide returnUrl on /logout


balazsorban44:
I have the following express implementation:

```js
// ...
express()
  .use(auth({
      required: false,
      idpLogout: true,
   }))
//...
```
I use the default `/login`, `/logout`, `/callback` routes.
`idpLogout` is set to `true`, because I want the user to be logged out from the Identity Provider as well. My problem is, that after logging out, they stay on the Identity Provider. 

**How can I pass a `returnUrl`, so the user comes back to the site?**
I tried putting this at the frontend ```<a href="/logout?returnUrl=http://example.com">Log out</a>```, but it did not help.

joshcanhelp:
@balazsorban44 - You can pass a `returnTo` parameter (instead of `returnUrl`) to that logout URL and you should get the behavior you're looking for:

https://github.com/auth0/express-openid-connect/blob/master/lib/context.js#L97

https://github.com/auth0/express-openid-connect/blob/master/API.md#response

Just make sure that URL is saved in your Auth0 Application in the **Allowed Logout URLs**. Give that a try and, if it's not doing what you expect, ping me here. 

That's a little buried, I'll try to surface that a bit better in our documentation here 👍

balazsorban44:
@joshcanhelp  You mean
```html
<a href="/logout?returnTo=http://localhost:8000">Log out</a>
```

or simply visiting `http://localhost:8000/logout?returnTo=http://localhost:8000`

should work? Now I am not using Auth0 but a custom Identity Provider, so I don't know if that is an issue, but there I have set the following:

![image](https://user-images.githubusercontent.com/18369201/69164420-44beb880-0af0-11ea-8684-8ad18f57ebff.png)

(Don't understand the difference between the two, so I set both 😊 Learning...)

joshcanhelp:
Ah, ok, you're not using Auth0. The library should work fine in most cases but logout is one place where we don't have a specification to rely on. I can help in a general sense but if there is custom handling around logout for your IdP, I'm not sure how much we can do here. 

So, setting `idpLogout` to `true` in the `auth()` configuration will clue the library in to look for a `end_session_endpoint` in your IdP's discovery document. This can be found by adding `.well-known/openid-configuration` to the end of your issuer base URL. Should look something like this:

https://auth0.auth0.com/.well-known/openid-configuration

Auth0 doesn't provide that key currently so we added another configuration option `auth0Logout` to handle Auth0 specifically:

[...]

PR #40, opened 2019-11-26
Fix post logout redirect, add config for default

balazsorban44:

### Description

When invoking the default `/logout`endpoint, we will try to attempt to get the returnURL from `req.query.returnTo`. In addition, I added the possibility to provide `post_logout_redirect_uri` in the config of `auth()`, like so:

```js
express()
//...
.use(auth({
    baseURL: "http://localhost:8000",
    post_logout_redirect_uri: "http://localhost:8000",
    required: false,
    idpLogout: true
  }))
// ...
```

@joshcanhelp I also did a change which I ask you to give a comment on. I moved the session destroy part _under_ the client.endSessionUrl() call, because `req.openid.tokens` becomes `undefined` after destroy. After this, my IdP redirects correctly as referred to in the issue referenced under. 

### References

This issue is an improvement after the discussion in #35.

### Testing

As of now I only tested it locally, but everything seems to work fine. If requested and/or the way I implemented it is accepted, I can try to create some tests in [logout.tests.js](https://github.com/auth0/express-openid-connect/blob/master/test/logout.tests.js)

### Checklist

Again, if my changes are accepted, I can document the changes.


balazsorban44:
ping @joshcanhelp :slightly_smiling_face:

lib/context.js
```diff
@@ -116,6 +110,13 @@ class ResponseContext {
         post_logout_redirect_uri: returnURL,
         id_token_hint: req.openid.tokens,
       });
+
+      if (typeof req.session.destroy === 'function') {
```
joshcanhelp: Great catch here! But can we move this and the redirect outside of the `try` block? It's only `client.endSessionUrl` that will throw an error so having the rest in there makes it a little less clear.

balazsorban44: Hmm. If `client.endSessionUrl` must come first and the redirect is dependent on it's result (`url`), the only way I can see it to achieve is using a `let`:
```js
const client = this._req.openid.client;
let url;
try {
  url = client.endSessionUrl({
    post_logout_redirect_uri: returnURL,
    id_token_hint: req.openid.tokens,
  });
} catch(err) {
  next(err);
}

if (typeof req.session.destroy === 'function') {
  req.session.destroy();
} else {
  req.session = null;
}
res.redirect(url);
```

Am I not seeing something, or you think this is the way to go?

lib/config.js
```diff
@@ -31,6 +31,7 @@ const paramsSchema = Joi.object().keys({
   errorOnRequiredAuth: Joi.boolean().optional().default(false),
   auth0Logout: Joi.boolean().optional().default(false),
   redirectUriPath: Joi.string().optional().default('/callback'),
+  post_logout_redirect_uri: Joi.string().optional(),
```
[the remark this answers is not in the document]
balazsorban44: Only used `post_logout_redirect_uri` instead of `postLogoutRedirectUri`, cause it is how it is referred to here:
https://github.com/auth0/express-openid-connect/blob/7050ffcd29470a396dbb2883c86349fb1ae24092/lib/context.js#L118
and here:
https://github.com/auth0/express-openid-connect/blob/7050ffcd29470a396dbb2883c86349fb1ae24092/lib/client.js#L57

But no problem to fix it. (I see I can also include it in the `index.d.ts` if we agree on the name.)

Also, not sure how to tell Joi to use `baseURL` as default:
`postLogoutRedirectUri: Joi.string().optional().default(????),`

And where to actually say that the param should default to baseURL. Somewhere in `buildAuthorizeParams`?

balazsorban44:
Moved some parts around a bit in the tests, hopefully it is OK with you. Moved the `before` call into a `setup` function, that takes some params to slightly change the result (able to modify auth params per test now, with a single setup).

Used this great article by Kent.C.Dodds: [Avoid Nesting when you're Testing
](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing)

Also moved repetitive strings to a single place (`mockData`, `authOParsedUrl`), so typos and forgetting to replace strings (eg.: version number in `pathname`) at all the places are out. (That part is only my own preference for the above mentioned reasons, might have been a bit fast with that, can change back if you wish)

lib/config.js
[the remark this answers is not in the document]
joshcanhelp: I see what you're saying. Maybe the better way to go about this is to follow the example of `redirectUriPath` and set that as a path, instead of a complete URI, and append the base URL when we use it. That would be more consistent and make a bit more sense as that should not be an external URL.

lib/context.js
joshcanhelp: I'm not wild about it either but I think it's more clear than packing all that work into a try block. If an error happens from this call, it's much more clear where it came from.

joshcanhelp:
RE: the test changes ... while I'm not under any delusions that the current way is the right way 😄, I'm a little worried about changing the whole testing paradigm in a single file. It's also not clear what tests were added for this functionality. Maybe we can address that approach in a different PR?

joshcanhelp:
@balazsorban44 - Are you able to wrap this one up in the near future? We're going to put out a release next week and it would be great if this was part of that. My main issue in merging this is the big changes in the logout tests. Would like to see primarily adding there, not a refactor. 

If you've moved on from the library, let me know and we'll handle it on our end. 

Thank you!

balazsorban44:
Hi, I am back after a long holiday. So to sum up, what is missing so this can be merged? (apart from fixing the conflicts).

joshcanhelp:
@balazsorban44 - Thanks for checking in. Main thing for me is to just see added tests for this functionality rather than a refactor of how the existing tests work. Also the comment about `redirectUriPath`, that makes more sense as a path since it will not be external.

balazsorban44:
OK, will fix next week :wink:

balazsorban44:
@joshcanhelp Fixed it, but because of the merged in `master` had some failing tests, now these tests also fail.

Both `postRedirectUri` and logout query string `returnTo` are relative paths now, which are concatenated with the `baseURL`.

balazsorban44:
We are close! (Sorry if it is a bigger task for such a small PR than you expected :sweat_smile: )

I know it is a bit back-and-forth, but I just looked up some docs before I made the name change:

[1. OpenID Connect Session Management](https://openid.net/specs/openid-connect-session-1_0.html) says:
> OPTIONAL. URL to which the RP is requesting that the End-User's User Agent be redirected after a logout has been performed. **The value MUST have been previously registered with the OP, either using the post_logout_redirect_uris Registration parameter or via another mechanism.** If supplied, the OP SHOULD honor this request following the logout.

[2. ASP.NET Core](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.openidconnectoptions.postlogoutredirecturi?view=aspnetcore-1.1) says:
> The uri where the user agent will be returned to after application is signed out from the identity provider. The redirect will happen after the SignedOutCallbackPath is invoked.
> **This URI can be out of the application's domain. By default it points to the root.**

Maybe none of these are relevant in our case, but I just wanted to clarify if you want to go that way, and really call it `postLogoutRedirectPath`, because although there is nothing wrong with it in my opinion, but it does not seem to appear in the OIDC spec in that form.  If you look at 1., it does say that you MUST register this URI beforehand. Using a relative path kind of does that, I guess...?

So I don't say I wouldn't change it to `postLogoutRedirectPath`, but I first would like to hear your opinion about this.

joshcanhelp:
Appreciate the research and thought on this @balazsorban44. And no worries about this becoming a long discussion. I would *much* rather talk this through completely and make a good choice than have to deal with pain points later. 

I typed out a bunch of stuff and ended up deleting it because the spec does not mention anything about the base **domain** of those post logout URIs. I was assuming that those needed to be on-site but that's not true according to the spec. As long as it's registered on the OP, then it's fine.

I'm going to run this by my team quickly but I'm now in agreement here about the `uri` part. Thank you again for starting the discussion!

One thing that would be nice here is to accept a path or a URI and append the base URL if it's a path. Does that sound reasonable?

balazsorban44:
Yes, supporting both may be the most reasonable decision. :) I will have a look at it.

balazsorban44:
Hmm. 🤔 Should full URIs be supported for `returnTo` as well? 🔒

index.d.ts
```diff
@@ -120,6 +120,19 @@ interface ConfigParams {
...
+     * or a valid URI to an external domain.
+     * The user will be redirected to this after a logout has been performed.
+     */
+    postLogoutRedirectUri?: string;
+
+    /**
+     * If you give a full URI in a `returnTo` parameter that does not start with the `baseURL`,
+     * then you must register it here.
+     */
+    postLogoutRedirectUris?: string[];
```
joshcanhelp: We don't need to confirm or validate the logout URL in the application, the auth server will, necessarily, do that for us. The `postLogoutRedirectUri` option should be the default parameter we send with the logout request. There's nothing wrong with passing along whatever comes through on `returnTo` as the auth server won't redirect back unless it was registered.

lib/config.js
```diff
@@ -113,6 +117,10 @@ module.exports.get = function(params) {
 
   config = paramsValidation.value;
 
+  if (!config.postLogoutRedirectUri) {
+    config.postLogoutRedirectUri = config.baseURL;
+  }
```
joshcanhelp: No need for this since we can now set the default.

lib/context.js
```diff
@@ -88,24 +89,51 @@ class ResponseContext {
...
+      return next(err);
+    }  
+    
+    let returnURL = returnTo || this._config.postLogoutRedirectUri;
+    
+    if (url.parse(returnURL).host === null) {
+      returnURL = urlJoin(this._config.baseURL, returnURL);
+    }
+    
+    if (!req.session || !req.openid) {
```
joshcanhelp: We removed the `req.session` requirement in 0.6.0 so we should not be checking that here. Use `req.isAuthenticated()`

lib/context.js
```diff
@@ -88,24 +89,51 @@ class ResponseContext {
...
     } catch(err) {
-      next(err);
+      return next(err);
+    }
+
+    if (typeof req.session.destroy === 'function') {
+      req.session.destroy();
+    } else {
+      req.session = null;
     }
```
joshcanhelp: Same here, no `req.session` handling needed.

test/logout.tests.js
```diff
@@ -44,7 +44,7 @@ describe('logout route', function() {
 
     it('should redirect to the base url', function() {
       assert.equal(logoutResponse.statusCode, 302);
-      assert.equal(logoutResponse.headers.location, 'https://example.org');
+      assert.equal(logoutResponse.headers.location, 'https://example.org/');
```
joshcanhelp: 🤔 why did this change?

balazsorban44: Either that, or

```diff
- postLogoutRedirectUri: Joi.string().uri({allowRelative: true}).optional().default('/')
+ postLogoutRedirectUri: Joi.string().uri({allowRelative: true}).optional().default('')
```

The default `'/'` gets appended to the baseURL.

Omitting `.default()` crashes the tests with:
```js
UnhandledPromiseRejectionWarning: TypeError [ERR_INVALID_ARG_TYPE]: The "url" argument must be of type string. Received type undefined
```

balazsorban44:
I reverted my `postLogoutRedirectUris` changes, and fixed the tests.

joshcanhelp:
@balazsorban44 - I tried to rebase master for you but your branch won't allow pushes. Can you resolve the conflicts here and push? If you're able to rebase the commits to cut down on the number there, that would be helpful as well.

PR #48, opened 2020-01-10
Generate test session during tests

joshcanhelp:
### Description

Generate test JWE cookie during tests to avoid broken tests.

### DIFF
--- a/test/appSession.tests.js
+++ b/test/appSession.tests.js
@@ -1,5 +1,6 @@
 const assert = require('chai').assert;
 const appSession = require('../lib/appSession');
+const sessionEncryption = require('./fixture/sessionEncryption');
 
 const defaultConfig = {
   name: 'identity',
@@ -8,10 +9,10 @@ const defaultConfig = {
   cookieOptions: {}
 };
 
-let req = {
+const req = {
   get: (key) => key
 };
-let res = {};
+const res = {};
 const next = () => true;
 
 describe('appSession', function() {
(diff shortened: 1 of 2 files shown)


test/fixture/sessionEncryption.js
```diff
@@ -0,0 +1,25 @@
+const { JWK, JWE } = require('jose');
```
joshcanhelp: Adapted from existing cookie encryption:

https://github.com/auth0/express-openid-connect/blob/master/lib/appSession.js

joshcanhelp:
We'll need to merge in #48 before these tests pass.

test/appSession.tests.js
```diff
@@ -54,19 +55,11 @@ describe('appSession', function() {
...
-    const result = appSessionMw(thisReq, res, next);
-
-    it('should call next', function() {
-      assert.ok(result);
-    });
+    const thisReq = {get: () => 'identity=' + sessionEncryption.encrypted};
 
     it('should set the identity on req', function() {
+      const result = appSessionMw(thisReq, res, next);
+      assert.ok(result);
```
lbalmaceda: If the line above is async (see `next` is being passed), how can you ensure the result will be set by when this line is reached? Unless I'm missing something

test/fixture/sessionEncryption.js
```diff
@@ -0,0 +1,25 @@
...
+const hkdf = require('futoin-hkdf');
+
+const deriveKey = (secret) => hkdf(secret, 32, { info: 'JWE CEK', hash: 'SHA-256' });
+const epoch = () => Date.now() / 1000 | 0;
```
lbalmaceda: Is this working? Or the pipe should be an OR? Also, I'd add parentheses for clarifying the scope.

```suggestion
const epoch = () => (Date.now() / 1000) || 0;
```

test/fixture/sessionEncryption.js
```diff
@@ -0,0 +1,25 @@
...
+const key = JWK.asKey(deriveKey('__test_secret__'));
+const payload = JSON.stringify({sub: '__test_sub__'});
+
+const encryptOpts = {
+  alg: 'dir',
+  enc: 'A256GCM',
+  zip: 'DEF',
+  uat: epoch(),
+  iat: epoch(),
+  exp: epoch() + ( 7 * 24 * 60 * 60 )
```
lbalmaceda: Given that epoch is a function, is there any chance that there might be a slight diff between these 3 different calls? and if that affects the token, wouldn't it be better to call the function once before defining the `encryptOpts` object?

Another nit one, I'd move `7 * 24 * 60 * 60` to a constant and give it a name, or at least add a comment giving the quick result e.g. `// 1 week in secs`

test/fixture/sessionEncryption.js
joshcanhelp: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#.7c_%28Bitwise_OR%29

joshcanhelp: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#.7c_%28Bitwise_OR%29

test/appSession.tests.js
joshcanhelp: The `next()` there isn't a good indication of whether that function is async or not. You would see `await` in front if it was.

PR #68, opened 2020-02-20
App session settings

joshcanhelp:
### Description

**This is a breaking change in configuration for all applications!**

This PR merges existing application session configuration into a single object. Specific changes:

- Add a required `appSession` configuration key that can be an object (described below) or `false` to skip the internally-generated cookie-based session.
- Change the default session name to `appSession` from `identity`
- Changed config key `appSessionSecret` to `appSession.secret`. If the `appSession` object is provided then this needs to have a string value 8 characters or greater, an array of string values 8 characters or greater, or an `APP_SESSION_SECRET` defined in `process.env`.
- Changed config key `appSessionName` to `appSession.name`.
- Changed config key `appSessionDuration` to `appSession.duration`.
- Merged config key `appSessionCookie` with the new `appSession` object. Cookie options are:
  - `cookieDomain`: Works the name as `appSessionCookie.domain` previously 
  - `cookieTransient`: Same functionality as `appSessionCookie.ephemeral` in v0.7.0, sets cookie expiration to `0`
  - `cookieHttpOnly`: Works the name as `appSessionCookie.httpOnly` previously 
  - `cookiePath`: Works the name as `appSessionCookie.path` previously 
  - `cookieSameSite`: Works the name as `appSessionCookie.sameSite` previously 
  - `cookieSecure`: Works the name as `appSessionCookie.secure` previously 
- TS: Removed `SessionCookieConfigParams` and added `AppSessionConfigParams`

An application configuration in 0.7.0 like this:

```js
// TS type here would be `SessionCookieConfigParams`
const appSessionCookieConfig = {
  domain: 'localhost',
  transient: false,
  httpOnly: true,
  path: '/',
  sameSite: 'Strict',
  secure: true
};

// TS type here would be `ConfigParams`
const oidcConfig: ConfigParams = {
  appSessionCookie: appSessionCookieConfig,
  appSessionSecret: process.env.APP_SESSION_SECRET,
  appSessionDuration: 234567890,
  appSessionName: 'customSession',
  issuerBaseURL: process.env.ISSUER_BASE_URL
};
```

Would now need to look like:

```js
// TS type here would be `AppSessionConfigParams`
const appSessionConfig = {
  secret: process.env.APP_SESSION_SECRET,
  duration: 234567890,
  name: 'customSession',
  cookieDomain: 'localhost',
  cookieTransient: false,
  cookieHttpOnly: true,
  cookiePath: '/',
  cookieSameSite: 'Strict',
  cookieSecure: true
};

// TS type here would still be `ConfigParams`
const oidcConfig: ConfigParams = {
  appSession: appSessionConfig,
  issuerBaseURL: process.env.ISSUER_BASE_URL
};
```

### Testing

### Checklist

### DIFF
--- a/lib/appSession.js
+++ b/lib/appSession.js
@@ -8,7 +8,7 @@ const hkdf = require('futoin-hkdf');
 const deriveKey = (secret) => hkdf(secret, 32, { info: 'JWE CEK', hash: 'SHA-256' });
 const epoch = () => Date.now() / 1000 | 0;
 
-module.exports = ({ name, secret, duration, cookieOptions = {} }) => {
+module.exports = (sessionConfig) => {
   let current;
 
   const COOKIES = Symbol('cookies');
@@ -17,11 +17,11 @@ module.exports = ({ name, secret, duration, cookieOptions = {} }) => {
 
   let keystore = new JWKS.KeyStore();
 
-  if (!Array.isArray(secret)) {
-    secret = [secret];
+  if (!Array.isArray(sessionConfig.secret)) {
+    sessionConfig.secret = [sessionConfig.secret];
   }
 
-  secret.forEach((secretString, i) => {
+  sessionConfig.secret.forEach((secretString, i) => {
     const key = JWK.asKey(deriveKey(secretString));
     if (i === 0) {
       current = key;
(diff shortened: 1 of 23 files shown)


lib/appSession.js
```diff
@@ -1,27 +1,27 @@
 const { strict: assert } = require('assert');
```
joshcanhelp: No intended change to session or encryption functionality.

PR #78, opened 2020-03-25
[SDK-1466] Allow to opt-out from sending SDK Telemetry - by setting `enableTelem…

adamjmcgrath:
### Description

Allow to opt-out from sending SDK Telemetry by setting `enableTelmetry: false`

This will prevent the library from sending the `Auth0-Client` header with the library and node version

### References

https://auth0team.atlassian.net/wiki/spaces/SDL/pages/588286390/express-openid-connect+SDK#Sensitive-Data-Exposure

### Testing

### Checklist

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -74,8 +74,8 @@ async function get(config) {
     // Allow configuration to override user agent header.
     {'User-Agent': `${pkg.name}/${pkg.version}`},
     httpOptions.headers || {},
-    // Do not allow overriding telemetry.
-    {'Auth0-Client': Buffer.from(JSON.stringify(telemetryHeader)).toString('base64')}
+    // Do not allow overriding telemetry, but allow it to be omitted.
+    config.enableTelemetry && {'Auth0-Client': Buffer.from(JSON.stringify(telemetryHeader)).toString('base64')}
   );
 
   custom.setHttpOptionsDefaults(httpOptions);
--- a/lib/client.js
+++ b/lib/client.js
@@ -78,7 +78,9 @@ async function get(config) {
     config.enableTelemetry && {'Auth0-Client': Buffer.from(JSON.stringify(telemetryHeader)).toString('base64')}
   );
 
-  custom.setHttpOptionsDefaults(httpOptions);
+  client[custom.http_options] = function(options) {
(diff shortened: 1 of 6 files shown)


test/client.tests.js
```diff
@@ -88,4 +88,27 @@ describe('client initialization', function() {
...
+      baseURL: 'https://example.org',
+      enableTelemetry: false
+    });
+
+    let client;
+    before(async function() {
+      client = await getClient(config);
+    });
+
+    it('should send the correct default headers', async function() {
```
damieng: Can we get a positive case too?

lib/client.js
```diff
@@ -74,11 +74,13 @@ async function get(config) {
...
     {'User-Agent': `${pkg.name}/${pkg.version}`},
     httpOptions.headers || {},
-    // Do not allow overriding telemetry.
-    {'Auth0-Client': Buffer.from(JSON.stringify(telemetryHeader)).toString('base64')}
+    // Do not allow overriding telemetry, but allow it to be omitted.
+    config.enableTelemetry && {'Auth0-Client': Buffer.from(JSON.stringify(telemetryHeader)).toString('base64')}
   );
 
-  custom.setHttpOptionsDefaults(httpOptions);
+  client[custom.http_options] = function(options) {
```
adamjmcgrath: Have changed how the default http options are set to setting them on a per request basis (https://github.com/panva/node-openid-client/blob/master/docs/README.md#customizing-individual-http-requests) to make this more testable.

This prevents an [issue in testing](https://circleci.com/gh/auth0/express-openid-connect/49), but it could also prevent an issue if there was a use case for creating multiple clients:

```
const clientA = await getClient({ httpOptions: { headers: { 'X-Custom-Header': 'bar' } } });
const clientB = await getClient({});

// Before this change, clientB will make requests with the `X-Custom-Header` even though this wasn't specified
```

This came up in the tests because I was testing for the absence of the `Auth0-Client` header when `enableTelemetry` was set to `false`.

This is because `setHttpOptionsDefaults` in `openid-client` mutates the defaults object which is scoped to the module level and can't be cleared https://github.com/panva/node-openid-client/blob/master/lib/helpers/request.js#L13 this makes it impossible to remove defaults when they're set.

PR #80, opened 2020-03-27
Change the default session duration to 1 day

adamjmcgrath:
### Description

Our security review suggested that 7 days for the app session cookie to expire is perhaps a little long for a default.

I had a look around, well just at `auth0-spa-js`, and noticed that uses 1 day.

Since the auth server handles the underlying session, it shouldn't noticeably effect user experience.

### References

https://auth0team.atlassian.net/wiki/spaces/SDL/pages/588286390/express-openid-connect+SDK#Security-Misconfiguration

### Testing

### Checklist

### DIFF
--- a/API.md
+++ b/API.md
@@ -23,7 +23,7 @@ Additional configuration keys that can be passed to `auth()` on initialization:
 
 - **`appSession`** - Object defining application session configuration. If this is set to `false`, the internal storage will not be used (see [this example](https://github.com/auth0/express-openid-connect/blob/master/EXAMPLES.md#4-custom-user-session-handling) for how to provide your own session mechanism). Otherwise, the `secret` key is required (see above).
   - **`appSession.secret`** - See the **Required Keys** section above.
-  - **`appSession.duration`** - Integer value, in seconds, for application session duration. Default is 7 days.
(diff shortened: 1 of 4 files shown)


lbalmaceda:
Yes! 🎉

YousefED:
@adamjmcgrath can you explain `Since the auth server handles the underlying session, it shouldn't noticeably effect user experience.`? Won't the value of 1 day mean all users have to re-login after 1 day?

adamjmcgrath:
@YousefED If the user is already logged in with their identity provider, the authorize call will just redirect them back to the application with the credentials, without requiring them to login.

YousefED:
> @YousefED If the user is already logged in with their identity provider, the authorize call will just redirect them back to the application with the credentials, without requiring them to login.

Thanks @adamjmcgrath. Makes sense. Does that also count for auth0 Database username / password provider? Which setting in Auth0 handles the session time on that side?

adamjmcgrath:
Hey @YousefED 

> Does that also count for auth0 Database username / password provider?
Yes, it should do

> Which setting in Auth0 handles the session time on that side?
In your tenant settings under "Advanced" - there's a section called "Log In Session Management"

PR #90, opened 2020-04-19
Update unauthorizedHandler.js

YousefED:
according to express, status can be set in both err.status or err.statuscode

https://expressjs.com/en/guide/error-handling.html

### DIFF
--- a/middleware/unauthorizedHandler.js
+++ b/middleware/unauthorizedHandler.js
@@ -7,7 +7,7 @@
 */
 module.exports = function() {
   return (err, req, res, next) => {
-    if (err.statusCode === 401) {
+    if (err.statusCode === 401 || err.status === 401) {
       return res.openid.login();
     }
     next(err);


middleware/unauthorizedHandler.js
```diff
@@ -7,7 +7,7 @@
 */
 module.exports = function() {
   return (err, req, res, next) => {
-    if (err.statusCode === 401) {
+    if (err.statusCode === 401 || err.status === 401) {
```
lbalmaceda, line 10: From the docs: 

> When an error is written, the following information is added to the response:
> - The res.statusCode is set from err.status (or err.statusCode). If this value is outside the 4xx or 5xx range, it will be set to 500.
> - The res.statusMessage is set according to the status code.

If the default error handler has already processed this by the time this middleware got called, does that mean that instead of this line with 2 inputs we can use `res.statusCode`? cc @stevehobbsdev

adamjmcgrath: > The res.statusCode is set from err.status (or err.statusCode)

Yep 👍 makes sense

adamjmcgrath: Would probably fix the codecov audit too

adamjmcgrath:
Closing this, feel free to reopen if you want to continue the conversation

PR #93, opened 2020-04-28
Handle `token_endpoint_auth_method` customization.

PelicanMaster:
### Description

Provide a way of customizing client's `token_endpoint_auth_method` using `tokenAuthMethod` option. It is then checked against Issuer's supported auth methods.
It was needed in my case because the default provided auth method was not compatible with my SSO provider. 

### Testing

Try the option `tokenAuthMethod` against any SSO.

### Checklist

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -84,10 +84,21 @@ async function get(config) {
     );
   }
 
+  const configTokenAuthMethod = config.tokenAuthMethod;
+  const issuerTokenAuthMethod = Array.isArray(issuer.token_endpoint_auth_methods_supported) ?
+    issuer.token_endpoint_auth_methods_supported : [];
+  if (configTokenAuthMethod && ! issuerTokenAuthMethod.includes(configTokenAuthMethod)) {
+    throw new Error(
+      `Token auth method "${configTokenAuthMethod}" is not supported by the issuer. ` +
+      `Supported token auth methods are "${issuerTokenAuthMethod.join('", "')}". `
+    );
+  }
+
   const client = new issuer.Client({
     client_id: config.clientID,
     client_secret: config.clientSecret,
     id_token_signed_response_alg: config.idTokenSigningAlg,
+    token_endpoint_auth_method: config.tokenAuthMethod,
   });
   applyHttpOptionsCustom(client);
   client[custom.clock_tolerance] = config.clockTolerance;
--- a/lib/client.js
+++ b/lib/client.js
@@ -84,16 +84,6 @@ async function get(config) {
     );
   }
 
-  const configTokenAuthMethod = config.tokenAuthMethod;
-  const issuerTokenAuthMethod = Array.isArray(issuer.token_endpoint_auth_methods_supported) ?
-    issuer.token_endpoint_auth_methods_supported : [];
-  if (configTokenAuthMethod && ! issuerTokenAuthMethod.includes(configTokenAuthMethod)) {
-    throw new Error(
-      `Token auth method "${configTokenAuthMethod}" is not supported by the issuer. ` +
-      `Supported token auth methods are "${issuerTokenAuthMethod.join('", "')}". `
-    );
-  }
-
   const client = new issuer.Client({
     client_id: config.clientID,
     client_secret: config.clientSecret,
--- a/lib/client.js
+++ b/lib/client.js
@@ -88,7 +88,7 @@ async function get(config) {
     client_id: config.clientID,
     client_secret: config.clientSecret,
     id_token_signed_response_alg: config.idTokenSigningAlg,
-    token_endpoint_auth_method: config.tokenAuthMethod,
+    token_endpoint_auth_method: config.clientAuthMethod,
   });
   applyHttpOptionsCustom(client);
   client[custom.clock_tolerance] = config.clockTolerance;
--- a/lib/config.js
+++ b/lib/config.js
@@ -148,13 +148,7 @@ const paramsSchema = Joi.object({
   })
     .default()
     .unknown(false),
-  loginPath: Joi.string().uri({relativeOnly: true}).optional().default('/login'),
-  logoutPath: Joi.string().uri({relativeOnly: true}).optional().default('/logout'),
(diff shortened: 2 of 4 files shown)


lib/client.js
```diff
@@ -48,10 +48,21 @@ async function get(config) {
...
+  const configTokenAuthMethod = config.tokenAuthMethod;
+  const issuerTokenAuthMethod = Array.isArray(issuer.token_endpoint_auth_methods_supported) ?
+    issuer.token_endpoint_auth_methods_supported : [];
+  if (configTokenAuthMethod && ! issuerTokenAuthMethod.includes(configTokenAuthMethod)) {
+    throw new Error(
+      `Token auth method "${configTokenAuthMethod}" is not supported by the issuer. ` +
+      `Supported token auth methods are "${issuerTokenAuthMethod.join('", "')}". `
+    );
+  }
+
```
panva, line 60: given that `token_endpoint_auth_methods_supported` is an optional attribute (which you handle) the throw is unexpected. With that, this whole diff can be removed.

lib/config.js
```diff
@@ -79,6 +79,7 @@ const paramsSchema = Joi.object({
   redirectUriPath: Joi.string().uri({relativeOnly: true}).optional().default('/callback'),
   required: Joi.alternatives([ Joi.function(), Joi.boolean()]).optional().default(true),
   routes: Joi.boolean().optional().default(true),
+  tokenAuthMethod: Joi.string().optional()
```
panva, line 82: we can add an enum here given that this abstraction over `openid-client` does not support some of the client authentication methods. I would suggest to enum `client_secret_basic` (default), `client_secret_post` and `none` to begin with.

API.md
```diff
@@ -49,6 +49,7 @@ Additional configuration keys that can be passed to `auth()` on initialization:
 - **`redirectUriPath`** - Relative path to the application callback to process the response from the authorization server. This value is combined with the `baseUrl` and sent to the authorize endpoint as the `redirectUri` parameter. Default is `/callback`.
 - **`required`** - Use a boolean value to require authentication for all routes. Pass a function instead to base this value on the request. Default is `true`.
 - **`routes`** - Boolean value to automatically install the login and logout routes. See [the examples](EXAMPLES.md) for more information on how this key is used. Default is `true`.
+- **`tokenAuthMethod`** - String value for the expected token auth method. Default is taken from Issuer supported methods.
```
panva, line 52:
```suggestion
- **`tokenAuthMethod`** - String value for the token endpoint authentication method. Default is `client_secret_basic`.
```

lib/client.js
PelicanMaster: I'm not sure I get it. The attribute itself is optional, but it doesn't mean it shouldn't be checked against the issuer ? It works the exact same way as `response_modes` or `response_types` ?

panva: Consider the following

```js
const config = {
  tokenAuthMethod: 'client_secret_basic'
};
const issuer = {};

const configTokenAuthMethod = config.tokenAuthMethod;
  const issuerTokenAuthMethod = Array.isArray(issuer.token_endpoint_auth_methods_supported) ?
    issuer.token_endpoint_auth_methods_supported : [];
  if (configTokenAuthMethod && ! issuerTokenAuthMethod.includes(configTokenAuthMethod)) {
    throw new Error(
      `Token auth method "${configTokenAuthMethod}" is not supported by the issuer. ` +
      `Supported token auth methods are "${issuerTokenAuthMethod.join('", "')}". `
    );
  }
```

It should be possible to configure an explicit auth method without the issuer having the optional property exposed. At the moment, this code will throw and it shouldn't.

Given that the issuer discovery property is optional, this code is completely unnecessary.

API.md
panva: The point here is that auth method should not be automatically selected based on an optional issuer property. There is a default used that can be overriden, that's fine, but there's no way to "select" the one that should be used from an unprioritized, unordered list.

lib/client.js
PelicanMaster: OK I get it. So is it better to completely remove this diff or check the attribute ONLY IF the issuer allows discovery ?

panva: completely remove

API.md
```diff
@@ -49,6 +49,7 @@ Additional configuration keys that can be passed to `auth()` on initialization:
 - **`redirectUriPath`** - Relative path to the application callback to process the response from the authorization server. This value is combined with the `baseUrl` and sent to the authorize endpoint as the `redirectUri` parameter. Default is `/callback`.
 - **`required`** - Use a boolean value to require authentication for all routes. Pass a function instead to base this value on the request. Default is `true`.
 - **`routes`** - Boolean value to automatically install the login and logout routes. See [the examples](EXAMPLES.md) for more information on how this key is used. Default is `true`.
+- **`tokenAuthMethod`** - String value for the token endpoint authentication method. Default is `client_secret_basic`.
```
lbalmaceda, line 52: Is it worth pointing here to an article where the accepted values are listed?

bnussbau:
can this be please be reviewed again and merged? Functionality of this PR works in my case.

rs22:
This is a useful addition that helped me in my application. I have squashed and rebased the changes onto the current master, where they work just fine.

@PelicanMaster could you update this PR to resolve the conflicts that currently prevent it from being merged? For reference, here's my branch where I already rebased your commits onto the current master: https://github.com/rs22/express-openid-connect/tree/token-endpoint-auth

PelicanMaster:
@rs22 I updated it but I can't test it anymore, let me know if I missed something.

rs22:
Thanks! I can confirm that it works for me.

panva:
@PelicanMaster I went ahead and integrated some internal feedback. Please let us know what you think.

PelicanMaster:
@panva This loks finne to me, thanks for adding unit tests.

panva (approved):

adamjmcgrath (approved):
Thanks @PelicanMaster and @panva - will run the ci tests locally (forked PRs don't run on our CI) and merge it

PR #95, opened 2020-05-03
Fix returnTo on Login

davidpatrick:
When no returnTo is passed into the login call, it should redirect to the URL that it was called from on GET Requests.

This is in line with documentation at https://github.com/auth0/express-openid-connect/blob/master/API.md#response

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -55,10 +55,17 @@ class ResponseContext {
     const config = this._config;
     const client = req.openid.client;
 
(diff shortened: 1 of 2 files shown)


jordanst3wart:
This has regressed with v2. It now redirects to the baseUrl, and not (what i think should be) the referrer.

https://github.com/auth0/express-openid-connect/blob/1d33b743fe9281c4cbf6f0d18f761b41502329de/middleware/auth.js#L42

PR #108, opened 2020-07-01
[SDK-1712] Test token set

adamjmcgrath:
### Description

Add some tests for the new token properties

Refactored the existing callback tests a little

### References

https://github.com/auth0/express-openid-connect/blob/beta-2/API.md#request

### Testing

### Checklist


test/fixture/server.js
```diff
@@ -26,9 +26,19 @@ module.exports.create = function (router, protect, path) {
...
+      refreshToken: req.oidc.refreshToken,
+      accessToken: req.oidc.accessToken,
+      idTokenClaims: req.oidc.idTokenClaims,
+    });
+  });
+
   if (protect) {
     app.get('/protected', protect, (req, res) => {
-      res.json(req.oidc.tokens);
+      res.json({});
```
adamjmcgrath, line 41: `req.oidc.tokens` prop doesn't exist, the tests just check the status code

test/callback.tests.js
```diff
@@ -0,0 +1,327 @@
+const assert = require('chai').assert;
```
adamjmcgrath, line 1: These tests are pretty much the same as `test/callback_route_form_post.tests.js` I just changed the name and reformatted them a little so they're a little more conventional (so I can run them from my IDE)

And added the last 2 tests

test/callback.tests.js
```diff
@@ -0,0 +1,327 @@
...
+        state: expectedDefaultState,
+        id_token: idToken,
+        code: 'jHkWEdUXMU1BwAsC4vtUsZwnNvTIxEl0z9K3vx5KF0Y',
+      }
+    });
+
+    assert.equal(tokens.isAuthenticated, true);
+    assert.equal(tokens.idToken, idToken);
+    assert.equal(tokens.refreshToken, '__test_refresh_token__');
+    assert.include(tokens.accessToken, {
```
panva, line 289: we ought to test also that

- `tokens.accessToken.expires_in` being a number (its required to be returned by all access token responses) that counts down every second
- `tokens.accessToken.isExpired` returns true/false

adamjmcgrath: 0c6814b

PR #109, opened 2020-07-03
[SDK-1715] Configuration and API updates

adamjmcgrath:
### Description

Add some tests, finish off TODOs and verify the changes to the Configuration

- added validation to check `openid` always in scope
- make sure `clientSecret` is required for `code id_token` as well as `code` with HS* signing algs
- `issuerBaseURL` can't be a hostname because `openid-client` `Issuer.discovery` only takes fully qualified URLs
- don't allow `null` or `fragment` in `response_type` because the callback handler doesn't accept `fragment`
- default to `query` for `response_type` `code` because you don't need to protect `code` with `form_post`
- add some extra tests to `config.test.js` and add invalid params tests to config so they're easier to find/use the same helpers/same format

### Testing

### Checklist


lib/config.js
```diff
@@ -50,23 +50,29 @@ const paramsSchema = Joi.object({
...
     response_type: Joi.string().optional().valid('id_token', 'code id_token', 'code').default('id_token'),
-    scope: Joi.string().optional().default('openid profile email'),
-    response_mode: Joi.alternatives([
-      Joi.string().optional(),
-      Joi.allow(null).optional()
-    ]).default((parent) => parent.response_type.includes('token') ? 'form_post' : undefined),
+    scope: Joi.string().optional().pattern(/\bopenid\b/, 'contains openid').default('openid profile email'),
+    response_mode: Joi.string().optional().when('response_type', {
+      is: 'code',
+      then: Joi.valid('query', 'form_post').default('query'),
```
panva, line 56: when code, leave `undefined` (i.e. no default) to omit the parameter from being sent, all servers are required to respond with a query automatically. For all other response types we send the value form_post explicitly because it's not the default to be used.

test/auth.tests.js
```diff
@@ -113,7 +113,7 @@ describe('auth', function () {
         assert.equal(parsed.query.client_id, '__test_client_id__');
         assert.equal(parsed.query.scope, 'openid profile email');
         assert.equal(parsed.query.response_type, 'code');
-        assert.equal(parsed.query.response_mode, undefined);
+        assert.equal(parsed.query.response_mode, 'query');
```
panva, line 116: as per the request to not default the mode to an explicit value this needs to be reverted too.

lib/config.js
```diff
@@ -50,23 +50,29 @@ const paramsSchema = Joi.object({
...
       then: Joi.string().required().messages({
-        'any.required': '"clientSecret" is required for response_type code'
-      }),
-      otherwise: Joi.when(
-        Joi.ref('idTokenSigningAlg', { adjust: (value) => value && value.substring(0, 2) === 'HS' }),
+        'any.required': '"clientSecret" is required for a response_type that includes code'
+      })
+    }
+  ).when(
+    Joi.ref('authorizationParams.response_type', { adjust: (value) => value && value.includes('id_token') }),
```
panva, line 71: `idTokenSigningAlg` being HS* is the requirement for needing a clientSecret, the response_type does not need to include `id_token` at all. ID Token will still be returned by the token endpoint with an HS signature even when response type is `code`. I believe the code before was correct and needed no changing (aside from the improved wording of the error message)

adamjmcgrath: Ah my bad, I thought that was a bug - will revert

lib/config.js
adamjmcgrath: 1da5afe

lib/config.js
adamjmcgrath: 193b93c

After our conversation I changed the behaviour to: `clientSecret` is required if `idTokenSigningAlg` is `HS*` regardless of `response_type`

lib/config.js
```diff
@@ -50,30 +50,30 @@ const paramsSchema = Joi.object({
...
-      )
+        'any.required': '"clientSecret" is required for a response_type that includes code'
+      })
+    }
+  ).when(
+    Joi.ref('idTokenSigningAlg', { adjust: (value) => value && value.startsWith('HS') }),
+    {
+      is: true,
+      then: Joi.string().required().messages({
+        'any.required': '"clientSecret" is required for ID tokens with HS algorithms'
```
panva, line 75:
```suggestion
        'any.required': '"clientSecret" is required for ID tokens with HMAC based algorithms'
```

PR #111, opened 2020-07-08
[SDK-1714] [SDK-1723] Session cookie checks

adamjmcgrath:
### Description

- added a couple of test cases for appSession
- reformatted appSession a little and made it use express mw
- updated the transientHelper tests a little

### Testing

### Checklist


lib/transientHandler.js
```diff
@@ -2,16 +2,16 @@ const { generators } = require('openid-client');
 const { JWKS, JWS, JWK } = require('jose');
 const { signing: deriveKey } = require('./hkdf');
 
-const HEADER = 'eyJhbGciOiJIUzI1NiIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19';
+const header = { alg: 'HS256', b64: false, crit: ['b64'] };
 const getPayload = (cookie, value) => Buffer.from(`${cookie}=${value}`);
 const flattenedJWSFromCookie = (cookie, value, signature) => ({
-  protected: HEADER,
+  protected:  Buffer.from(JSON.stringify(header)).toString('base64'),
```
panva, line 8: Well, this only works because there are no specific base64url chars in the output.

This would be the correct way of encoding it that works if we'd change the header.
```js
base64url.encode(JSON.stringify({ alg: 'HS256', b64: false, crit: ['b64'] }))
```

I didn't want to pull in base64url as a dependency just for this so i hardcoded the header value instead.

panva, line 8:
```suggestion
  protected:  Buffer.from(JSON.stringify(header)).toString('base64').replace(/=/g, '').replace(/\+/g, '-').replace(/\//g, '_'),
```

adamjmcgrath: Ah ok, I thought there might be a reason

PR #112, opened 2020-07-13
Add Prettier for style formatting

adamjmcgrath:
### Description

Adding prettier 050d217
Running prettier (big diff, but all formatting) ac1a780


.eslintrc.json
```diff
@@ -1,35 +1,16 @@
...
-        "indent": [
-            "error",
-            2,
-            { "SwitchCase": 1 }
-        ],
-        "linebreak-style": [
-            "error",
-            "unix"
-        ],
-        "quotes": [
```
adamjmcgrath, line 20: Don't need formatting rules as prettier will handle this

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
```
adamjmcgrath, line 1: Same config as auth0.js, spa js, auth0-react etc..

PR #113, opened 2020-07-14
[SDK-1716] Add tests for claim* MW

adamjmcgrath:
### Description

Added some tests for the `requiresAuth.claim*` MW
Changed the behaviour of `checkClaim` to return `true` for allowed, `false` for unauthorized
Doing docs in another PR

### Testing

### Checklist


middleware/requiresAuth.js
```diff
@@ -133,7 +126,7 @@ module.exports.custom = function custom(func) {
 
     const { idTokenClaims } = req.oidc;
 
-    return func(req, idTokenClaims);
+    return !func(req, idTokenClaims);
```
adamjmcgrath, line 129: I thought `claimCheck(() => true)` should mean allow, `claimCheck(() => false)` should mean 401

PR #120, opened 2020-08-04
Add TROUBLESHOOTING and update debug logging

adamjmcgrath:
### Description

Added a TROUBLESHOOTING.md and updated some `debug` statements

image

### Checklist

### DIFF
--- a/lib/appSession.js
+++ b/lib/appSession.js
@@ -9,7 +9,7 @@ const onHeaders = require('on-headers');
 const cookie = require('cookie');
 const COOKIES = require('./cookies');
 const { encryption: deriveKey } = require('./hkdf');
-const debug = require('./debug');
+const debug = require('./debug')('appSession');
 
 const epoch = () => (Date.now() / 1000) | 0;
 const CHUNK_BYTE_SIZE = 4000;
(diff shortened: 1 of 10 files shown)


lib/appSession.js
```diff
@@ -9,7 +9,7 @@ const onHeaders = require('on-headers');
 const cookie = require('cookie');
 const COOKIES = require('./cookies');
 const { encryption: deriveKey } = require('./hkdf');
-const debug = require('./debug');
+const debug = require('./debug')('appSession');
```
adamjmcgrath, line 12: I removed the levels because they didn't work like logging levels (`warn:*` wouldn't include `info:*` etc).

But I liked how express split the logs up by filename (https://github.com/expressjs/express/blob/master/lib/application.js#L21)

lib/context.js
```diff
@@ -222,19 +223,22 @@ class ResponseContext {
...
 
     let returnURL = params.returnTo || config.routes.postLogoutRedirectUri;
+    debug('req.oidc.logout() with return url: %s', returnURL);
 
     if (url.parse(returnURL).host === null) {
       returnURL = urlJoin(config.baseURL, returnURL);
     }
 
     if (!req.oidc.isAuthenticated()) {
+      debug('user already logged out, redirecting to %s', returnURL);
```
panva, line 233:
```suggestion
      debug('end-user already logged out, redirecting to %s', returnURL);
```

PR #124, opened 2020-08-10
[SDK-1877] Add refresh method to access token

adamjmcgrath:
### Description

You should be able to use the `openid-client` to refresh an access token.

```js
let accessToken = req.oidc.accessToken;
if (accessToken.isExpired()) {
  accessToken = await accessToken.refresh();
}
```

Also, add `fetchUserInfo` for completeness

```js
const userInfo = await req.oidc.fetchUserInfo();
```

### References

https://github.com/panva/node-openid-client/blob/master/docs/README.md#clientrefreshrefreshtoken-extras
https://github.com/panva/node-openid-client/blob/master/docs/README.md#clientuserinfoaccesstoken-options

### Testing

### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,10 @@
 // Type definitions for express-openid-connect
 
-import { AuthorizationParameters, IdTokenClaims } from 'openid-client';
+import {
+  AuthorizationParameters,
+  IdTokenClaims,
+  RefreshExtras,
+} from 'openid-client';
 import { Request, Response, RequestHandler } from 'express';
 
 /**
@@ -71,27 +75,7 @@ interface RequestContext {
    *
    * See: https://auth0.com/docs/protocols/oidc#access-tokens
    */
-  accessToken?: {
-    /**
-     * The access token itself, can be an opaque string, JWT, or non-JWT token.
-     */
-    access_token: string;
-
-    /**
-     * The type of access token, Usually "Bearer".
-     */
-    token_type: string;
-
-    /**
-     * Number of seconds until the access token expires.
-     */
-    expires_in: number;
-
-    /**
-     * Returns `true` if the access_token has expired.
-     */
-    isExpired: () => boolean;
-  };
+  accessToken?: AccessToken;
 
   /**
    * Credentials that can be used to refresh an access token.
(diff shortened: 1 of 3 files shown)


lib/context.js
```diff
@@ -15,6 +15,20 @@ function isExpired() {
   return tokenSet.call(this).expired();
 }
 
+async function refresh(opts) {
+  let { config, req } = weakRef(this);
+  const client = await getClient(config);
+  const oldTokenSet = tokenSet.call(this);
+  const newTokenSet = await client.refresh(oldTokenSet, opts);
```
panva, line 22: It is entirely possible that a refresh() will only yield the access token and no refresh_token. We should look for that specific scenario and keep the old refresh token value.

index.d.ts
```diff
@@ -110,6 +95,27 @@ interface RequestContext {
...
+   * ```
+   *
+   * @param options Options for the UserInfo request.
+   */
+  fetchUserInfo(options?: {
+    verb?: 'GET' | 'POST';
+    via?: 'header' | 'body' | 'query';
+    tokenType?: string;
+    params?: object;
+  }): Promise<UserinfoResponse>;
```
panva, line 118: my recommendation: don't expose any of these options. The default is the REQUIRED to implement for all providers that support the userinfo endpoint.

lib/context.js
```diff
@@ -112,6 +127,13 @@ class RequestContext {
       return undefined;
     }
   }
+
+  async fetchUserInfo(opts) {
```
panva, line 131: my recommendation: don't expose the options. The default is the REQUIRED to implement for all providers that support the userinfo endpoint.

index.d.ts
```diff
@@ -454,6 +460,42 @@ interface SessionConfigParams {
...
+   * ```js
+   * let accessToken = req.oidc.accessToken;
+   * if (accessToken.isExpired()) {
+   *   accessToken = await accessToken.refresh();
+   * }
+   * ```
+   *
+   * @param opts Add extra parameters to the Token Endpoint Request and/or Client Authentication JWT Assertion
+   */
+  refresh(opts?: RefreshExtras): Promise<AccessToken>;
```
panva, line 496: omit the options.

lib/context.js
```diff
@@ -15,6 +15,20 @@ function isExpired() {
   return tokenSet.call(this).expired();
 }
 
+async function refresh(opts) {
```
panva, line 18: omit the options.

lib/context.js
adamjmcgrath: Good point, thanks!

PR #127, opened 2020-08-19
[SDK-1876] [SDK-1726] Add samples and smoke tests

adamjmcgrath:
### Description

Mostly samples an tests:

- Update ./EXAMPLES.md (removed examples for custom storage and state - these will be handled in later work)
- Created running apps for each example that can be run against auth server specified with env vars or agains mock auth server.
- Created e2e smoke tests for each sample against mock auth server
- Also fixed a bug where refreshed ATs were not being persisted in the session

### Testing

### Checklist

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,23 @@
+# Examples
+
+To run an example `npm run start:example -- "name of example"`. Eg to run the basic example at `./basic.js`:
+
+```shell script
+$ npm run start:example -- basic
+```
+
+To run the example against your authorization server add the following items to your `./examples/.env`
+
+```shell script
+# For the example app
+PORT=3000
+# For the auth config
+ISSUER_BASE_URL=https://YOUR_DOMAIN
+CLIENT_ID=YOUR_CLIENT_ID
+BASE_URL=https://YOUR_APPLICATION_ROOT_URL
+SECRET=LONG_RANDOM_VALUE
+# For response_type values that include 'code'
+CLIENT_SECRET=YOUR_CLIENT_SECRET
+```
+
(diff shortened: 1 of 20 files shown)


lib/context.js
```diff
@@ -20,15 +20,21 @@ async function refresh() {
...
   const newTokenSet = await client.refresh(oldTokenSet);
-  // If no new refresh token assume the current refresh token is valid.
-  if (!newTokenSet.refresh_token) {
-    newTokenSet.refresh_token = oldTokenSet.refresh_token;
-  }
 
-  // Update the session's tokenSet
+  // Update the session
   const session = req[config.session.name];
+  Object.assign(session, {
```
adamjmcgrath, line 26: Fixed issue where refreshed ATs were not being persisted in the session.

I'm now updating `req[config.session.name]` rather than the `cachedTokenSet`

Test case https://github.com/auth0/express-openid-connect/pull/127/files#diff-1840f764728973b9eb715f3e751c1978

examples/README.md
```diff
@@ -0,0 +1,23 @@
+# Examples
+
+To run an example `npm run start:example -- "name of example"`. Eg to run the basic example at `./basic.js`:
+
+```shell script
+$ npm run start:example -- basic
+```
```
stevehobbsdev, line 7: I recommend listing the available examples here. I know they're listed out somewhat on the EXAMPLES.md page, but would be nice to have a concise reference somewhere here.

stevehobbsdev: Actually I can see how the "basic" example would be ok, in that the name is the same as the filename. Still, interested in your thoughts.

stevehobbsdev (approved):
LGTM, just a small thing raised but not blocking.

examples/README.md
adamjmcgrath: It's probably one of those things that would go out of date, also when you're looking at the `examples/README.md` you'll usually have the latest list above:

image

PR #130, opened 2020-08-25
Make the test clearer that discovery alg is ignored

adamjmcgrath:
### Description

Update test to make it clearer that if the discovery doc is compromised the `idTokenSigningAlg` can not be overridden from the client configuration options.

And add a test to show "none" is not allowed for `idTokenSigningAlg` config

### DIFF
--- a/test/config.tests.js
+++ b/test/config.tests.js
@@ -370,6 +370,18 @@ describe('get config', () => {
     );
   });
 
+  it('should not allow "none" for idTokenSigningAlg', () => {
+    const config = {
+      ...defaultConfig,
+      idTokenSigningAlg: 'none',
+    };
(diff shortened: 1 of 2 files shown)


test/config.tests.js
```diff
@@ -370,6 +370,18 @@ describe('get config', () => {
     );
   });
 
+  it('should not allow "none" for idTokenSigningAlg', () => {
```
lbalmaceda, line 373: the test above is for the alg received from the .well-known and this one is for the configuration/options passed by the user. Is that correct? If so, looks good, and thanks for updating these.

adamjmcgrath: yep - I thought I'd add one for the configuration as well

PR #134, opened 2020-09-09
beta-2: update dependencies

panva:
This updates

- `jose` to 2.x and accomodates its breaking changes (actual API changes)
- `openid-client` to 4.x and accomodates its breaking changes (node engines change)

Decomposing the final semver range `^10.19.0 || >=12.0.0 < 13 || >=13.7.0 < 14 || >= 14.2.0`

- `^10.19.0` - minimal for updated `openid-client > got` dependency
- `>= 12.0.0` - general requirement
- `< 13 || >=13.7.0` - everything 13.x until 13.7.0 had a bug in its ESM implementation which makes requiring both jose and openid-client under those node.js versions fail
- `< 14 || >= 14.2.0` - Node.js core networking module had a snafuu, let's call it that, that made several http clients (including `openid-client > got`) simply not work.


adamjmcgrath:
lgtm

Issue #151, opened 2020-10-22
Support for WebSocket upgrade authentication


edwardb96:
### Describe the problem you'd like to have solved
Assuming a user has already logged in via the usual means, it would be handy to be able to use the session cookie to authenticate users during the WebSocket upgrade request (assuming websocket server and http server running on the same domain).

I don't think it's currently easy/possible to use this library to implement an authenticate function suitable for the upgrade request as in [this](https://github.com/websockets/ws#client-authentication) code sample.

I'm relatively new to how openid connect works so I guess there could be some technical reason as to why this isn't possible.

### Describe the ideal solution
An authenticate function with a matching signature or something you could just import or have optionally returned from calling `auth` (or a similar function) would be ideal. Failing that, the ability to call a function to validate the session cookie and get the `oidc`/`oidc.user` objects/profile would probably do.

### Hacking it together (not yet working)
I browsed through the code and started hacking something together which returned the `appSession` instance created [here](https://github.com/auth0/express-openid-connect/blob/5949c646b5e4c8be56b597bb0bbe36a3865b666c/middleware/auth.js#L32) in an object along with the router. 

I also added an if [here](https://github.com/auth0/express-openid-connect/blob/5949c646b5e4c8be56b597bb0bbe36a3865b666c/lib/appSession.js#L236) to cope with a `null` response object.

My plan was then to call it like this

```ts
import { get as getConfig } from 'express-openid-connect/lib/config';
import { RequestContext } from 'express-openid-connect/lib/context';
import { auth, requiresAuth } from 'express-openid-connect';

const authConfig = getConfig(authParams);
const { router, appSessionMiddleware } = (auth(authParams) as any);
// ...
server.on('upgrade', async (req, socket, head) => {
  appSessionMiddleware(req, null, async req => {
    try {
      const oidc = new RequestContext(authConfig, req, null, () => console.log("next was called"))
      const userInfo = await oidc.fetchUserInfo();
      console.log("Got user info %s", userInfo);
    } catch(e) {
      console.log("Got err %s", e);
    }
  }
}
```

I then hit a problem with [this line](https://github.com/auth0/express-openid-connect/blob/5949c646b5e4c8be56b597bb0bbe36a3865b666c/lib/appSession.js#L156) and remembered that the req object express gives you is different to the req/socket object you get from http, however it's still possible to get the cookie values from `signedCookies`, I think it just has a slightly different interface.

### PassportJS and Cookie Parser
[...]

PR #152, opened 2020-10-24
WIP: Add initial implementation of websocket authentication

edwardb96:
***This pull request is a work in progress and is not ready to merge quite yet***

### Description
Adds the ability to authenticate WebSocket upgrade requests using a pre-existing session cookie as requested in #151.

Can be enabled by setting the new `webSocket` config option to true. This calls an overload of `auth` which returns both the express `Router` as well as a function with type `(req: http.IncomingMessage,  next: (err: unknown, oidc: RequestContext) => void;` which can be used as in [this](https://github.com/websockets/ws#client-authentication) code sample to authenticate web socket connections when the HTTP upgrade request is recieved.

### References
Issue #151 
https://github.com/websockets/ws#client-authentication

> Also include details of the environment this PR was developed in (language/platform/browser version).
- This PR was developed and tested on Fedora 32 Linux using yarn 1.22.10 with Node v12.18.3 with Firefox 80.1.

- [x] This change adds test coverage for new/changed/fixed functionality - *minimal at present*

### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -6,6 +6,7 @@ import {
   UserinfoResponse,
 } from 'openid-client';
 import { Request, Response, RequestHandler } from 'express';
+import { IncomingMessage } from 'http';
 
 /**
  * The Express.js Request with `oidc` context added by the `auth` middleware.
@@ -522,7 +523,30 @@ interface AccessToken {
  *  app.listen(3000, () => console.log('listening at http://localhost:3000'))
  * ```
  */
-export function auth(params?: ConfigParams): RequestHandler;
+export function auth(params?: ConfigParamsWithoutWebSocket): RequestHandler;
+export function auth(params: ConfigParamWithWebSocket): Authenticators;
+
+interface ConfigParamsWithoutWebSocket extends ConfigParams {
+  webSocket?: false;
+}
+
+interface ConfigParamWithWebSocket extends ConfigParams {
+  webSocket: true;
+}
+
+type WebSocketAuthenticatorCallback = (
+  err: unknown,
+  oidc: RequestContext
+) => void;
+type WebSocketAuthenticator = (
+  req: IncomingMessage,
+  next: WebSocketAuthenticatorCallback
+) => void;
+
+interface Authenticators {
+  expressAuthRouter: RequestHandler;
+  webSocketAuthenticate: WebsocketAuthenticator;
+}
 
 /**
  * Set {@link ConfigParams.authRequired authRequired} to `false` then require authentication
--- a/lib/appSession.js
+++ b/lib/appSession.js
@@ -140,7 +140,93 @@ module.exports = (config) => {
     }
   }
 
-  return (req, res, next) => {
+  function readExistingSessionCiphertext(cookies) {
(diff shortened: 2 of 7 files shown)


edwardb96:
Hi, 

Sorry, I'm back again. 

Thanks for your suggestions, I had high hopes for the code sample you sent - thought it would be a copy+paste job so finished the other features I was working on before coming back to this.

The idea of upgrading the request did occur to me originally but I couldn't find an express API for upgrading the request object.

I tried out your code sample this morning but despite everything looking ok from the server side (no-crashes, successful authentication) on the client side the upgrade doesn't seem to work. 

I get an error event and a the connection gets closed with code 1006 and an empty 'reason' string. 
This happens even with the line `res.on('finish', () => res.socket.destroy());` commented out <sup>1</sup>. 
It looks as if the client also gets sent a 404 for the websocket server address (`wss://localhost/api`) in my case.

I haven't yet got a full explanation for why this doesn't work but my best guess from some debug logging and reading some of the source from the expressjs repo would be that calling `app(req, res)` is not only monkey patching the `req` object but also trying to handle it like an ordinary http request and since, as far as express is concerned, the route `wss://localhost/api` doesn't exist so it sends back a http 404. This then messes up the websocket upgrade handshake? - more details to follow later if I can find them after some more thorough debugging.

This brings me on to express-ws, I came across this with my old PassportJS setup, and tried to use it for a while but kept hitting a bug<sup>2</sup> in an old version of the underlying ws library. I believe [`express-ws`](https://github.com/HenningM/express-ws) is an unofficially dead project, it hasn't had a commit or release for over two years now, this feels like a non-starter.

Finally, access token jwts, I have a very limited understanding as to what these are or how they work. If they can be used with the free tier of Auth0 then I'm happy to learn more and give them a try. It does feel a little annoying that despite using what should be the easy way to do authentication, the fact that I need WebSockets has made it take up significantly more time due to a lack of working library support, especially when the cookies needed are already trivially accessible.

I'll do some reading and decide whether to try jwts or continue debugging. It would be useful to know if you have any further thoughts on this. I understand the concerns regarding scope but wonder if it might be possible to restructure some of the internals so that it's possible/easier to implement websocket authentication from another package without having to maintain a complete fork?

<sup>1</sup> I think, this is necessary because otherwise the connection seems to get dropped immediately after the upgrade when the ServerResponse goes out of scope.
<sup>2</sup> Sending messages when the socket wasn't in the readyState or something - it was months ago now.

edwardb96:
I've managed to put something together using the jsonwebtoken library to sign/create and later verify a token which contains the user id but which isn't the idToken mentioned above. The token is passed as a query string parameter in the WebSocket connection URL so it is available as part of the upgrade request. I've made sure to give the token a short expiry time, and not included any sensitive user information in the token, so if the token is decoded by a third party it is no use and only valid for a few seconds. My understanding is that this should be reasonably secure for my purposes.

adamjmcgrath:
Hey @edwardb96 - apologies, I missed this the first time around.

I've put a working example of my earlier suggestion https://github.com/auth0/express-openid-connect/pull/152#issuecomment-719682819 here https://github.com/adamjmcgrath/eoidc-testing-example/tree/ws

PR #159, opened 2020-11-12
[SDK-2155] Default cookie.secure config to the protocol of baseURL

### LINKED ISSUE #145: BadRequestError: checks.state argument is missing
### Description

Sometimes while testing this I get:
```
BadRequestError: checks.state argument is missing
    at <folder>/node_modules/express-openid-connect/middleware/auth.js:105:31
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```
This happens randomly after logging in.

### Reproduction

It happens especially often at the first login, but sometimes just randomly after logging in. My code:
```js
require('dotenv').config()
const express = require('express')
const app = express()
const { auth } = require('express-openid-connect')
const config = {
  authRequired: false,
  auth0Logout: true,
  issuerBaseURL: process.env.ISSUER_BASE_URL,
  secret: process.env.SECRET,
}
app.use(auth(config))
app.get('/', (req, res) => {
  res.send(req.oidc.isAuthenticated() ? 'Logged in. Welcome!' : 'Logged out. <a href="/login">Login?</a>')
})
app.listen(3000)
```

### Environment

- **Version of this library used:** 2.0.0
- **Version of the platform or framework used, if applicable:** Express version 4.17.1
- **Other relevant versions (language, server software, OS, browser):** Node.js version 12.18.4 installed with nvm, macOS 10.15.6, Chrome 85.0.4183.121
- **Other modules/plugins/libraries that might be involved:** dotenv version 8.2.0

adamjmcgrath:
### Description

Default the `cookie.secure` config to the protocol of the `baseUrl` config - rather than defaulting it to the protocol of the `req.url` (which relies on the developer to set trust proxy when behind a proxy)

And a couple of other things that should improve the DX around this setting:

- if you override `cookie.secure` to `true` and `baseUrl` is `http` - throw an error, because this will not work (your secure cookies won't be read over http)
- if you override `cookie.secure` to `false` and `baseUrl` is `https` - warn the user that the cookie will not be set with the Secure attribute
- If `baseUrl` is `http` and `response_mode` is `form_post` - warn the use that this will only work for <2min logins (have also added an FAQ for this)

### References

Chrome 2 min exception for Lax+POST https://www.chromium.org/updates/same-site
fixes #145 

### Testing

### Checklist

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,0 +1,13 @@
+# FAQs
+
+## I'm getting the warning "Using 'form_post' for response_mode may cause issues for you logging in over http" on localhost
+
+If you use `form_post` response mode (the default for this library) you are relying on a cross-site POST request with cookies - these will only be sent if they specify `SameSite=None; Secure` properties.
+
+If your server is running over `http`, your cookie will not be sent with the `Secure` attribute and, under the new SameSite behavior, will be blocked.
+
+However, there is the "Lax+POST" [special exception](https://www.chromestatus.com/feature/5088147346030592) that Chrome makes for such cookies for the first 2 minutes after they are created which means that your logins will work in Chrome over `http` as long as they take less than 2 minutes, longer logins will fail. This special exception will be phased out in future Chrome releases.
+
+This should not be an issue on production, because your application will be running on `https`
+
+To resolve this, you could use the [code flow](https://auth0.com/docs/flows/authorization-code-flow) or [run your local development server over https](https://auth0.com/docs/libraries/secure-local-development)
--- a/FAQ.md
+++ b/FAQ.md
@@ -2,12 +2,12 @@
 
 ## I'm getting the warning "Using 'form_post' for response_mode may cause issues for you logging in over http" on localhost
 
-If you use `form_post` response mode (the default for this library) you are relying on a cross-site POST request with cookies - these will only be sent if they specify `SameSite=None; Secure` properties.
+If you use `form_post` response mode (the default for this library) you are relying on a cross-site POST request with cookies - these will only be attached to the POST request if they were set with `SameSite=None; Secure` properties.
 
-If your server is running over `http`, your cookie will not be sent with the `Secure` attribute and, under the new SameSite behavior, will be blocked.
+If your server is running over `http:` protocol, your cookie with the `Secure` property will not be attached under current browser SameSite behavior.
 
-However, there is [an exception](<(https://www.chromestatus.com/feature/5088147346030592)>) for "Lax+POST" that Chrome makes for such cookies for the first 2 minutes after they are created. This means that your logins will work in Chrome over `http` as long as they take less than 2 minutes, longer logins will fail. This special exception will be phased out in future Chrome releases.
+However, there is [an exception](<(https://www.chromestatus.com/feature/5088147346030592)>) for "Lax+POST" that Chrome makes for such cookies for the first 2 minutes after they are stored. This means that your login requests will work in Chrome over `http` as long as the end-user takes less than 2 minutes to authenticate, otherwise it will fail. This special exception will be phased out in future Chrome releases.
 
-This should not be an issue on production, because your application will be running on `https`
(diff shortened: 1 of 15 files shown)


FAQ.md
```diff
@@ -0,0 +1,13 @@
+# FAQs
+
+## I'm getting the warning "Using 'form_post' for response_mode may cause issues for you logging in over http" on localhost
+
+If you use `form_post` response mode (the default for this library) you are relying on a cross-site POST request with cookies - these will only be sent if they specify `SameSite=None; Secure` properties.
```
panva, line 5:
```suggestion
If you use `form_post` response mode (the default for this library) you are relying on a cross-site POST request with cookies - these will only be attached to the POST request if they were set with `SameSite=None; Secure` properties.
```

FAQ.md
```diff
@@ -0,0 +1,13 @@
...
+
+If your server is running over `http`, your cookie will not be sent with the `Secure` attribute and, under the new SameSite behavior, will be blocked.
```
panva, line 7:
```suggestion
If your server is running over `http:` protocol, your cookie with the `Secure` property will not be attached under current browser SameSite behavior.
```

FAQ.md
```diff
@@ -0,0 +1,13 @@
...
+
+However, there is [an exception](<(https://www.chromestatus.com/feature/5088147346030592)>) for "Lax+POST" that Chrome makes for such cookies for the first 2 minutes after they are created. This means that your logins will work in Chrome over `http` as long as they take less than 2 minutes, longer logins will fail. This special exception will be phased out in future Chrome releases.
```
panva, line 9:
```suggestion
However, there is [an exception](<(https://www.chromestatus.com/feature/5088147346030592)>) for "Lax+POST" that Chrome makes for such cookies for the first 2 minutes after they are stored. This means that your login requests will work in Chrome over `http` as long as the end-user takes less than 2 minutes to authenticate, otherwise it will fail. This special exception will be phased out in future Chrome releases.
```

FAQ.md
```diff
@@ -0,0 +1,13 @@
...
+
+This should not be an issue on production, because your application will be running on `https`
```
panva, line 11:
```suggestion
This should not be an issue in production, because your application will be running over `https`
```

FAQ.md
```diff
@@ -0,0 +1,13 @@
...
+
+To resolve this, you could [run your local development server over https](https://auth0.com/docs/libraries/secure-local-development) or use the [code flow](https://auth0.com/docs/flows/authorization-code-flow).
```
panva, line 13:
```suggestion
To resolve this, you should [run your local development server over https](https://auth0.com/docs/libraries/secure-local-development).
```

lib/config.js
```diff
@@ -45,7 +47,23 @@ const paramsSchema = Joi.object({
...
+            if (!value) warn('insecure.cookie');
+            return value;
+          })
+          .messages({
+            'insecure.cookie':
+              "Setting your cookie to insecure when over https is not recommended, I hope you know what you're doing.",
+          }),
+        otherwise: Joi.boolean().valid(false).default(false).messages({
+          'any.only':
+            'Cookies set with the `Secure` property wont be read over http',
```
panva, line 64:
```suggestion
            'Cookies set with the `Secure` property wont be attached to http requests',
```

test/config.tests.js
```diff
@@ -177,6 +198,40 @@ describe('get config', () => {
...
   });
 
+  it('should fail when the baseURL is http and cookie is secure', function () {
+    assert.throws(() => {
+      getConfig({
+        ...defaultConfig,
+        baseURL: 'http://example.com',
+        session: { cookie: { secure: true } },
+      });
+    }, 'Cookies set with the `Secure` property wont be read over http');
```
panva, line 208:
```suggestion
    }, 'Cookies set with the `Secure` property wont be attached to http requests');
```

panva (commented):
LGTM, i'm leaving some text suggestions.

panva (approved):

PR #168, opened 2020-12-14
Move transient cookies into one cookie [SDK-1908]

davidpatrick:
Currently all transient cookies (state, nonce, max_age and code_verifier) are stored individually.  This refactors these values to be stored in a single cookie.

### DIFF
--- a/middleware/auth.js
+++ b/middleware/auth.js
@@ -83,31 +83,25 @@ module.exports = function (params) {
         try {
           const redirectUri = res.oidc.getRedirectUri();
 
-          let expectedState;
           let session;
 
           try {
             const callbackParams = client.callbackParams(req);
-            expectedState = transient.getOnce('state', req, res);
-            const max_age = parseInt(
-              transient.getOnce('max_age', req, res),
-              10
-            );
-            const code_verifier = transient.getOnce('code_verifier', req, res);
-            const nonce = transient.getOnce('nonce', req, res);
+            const authVerification = transient.getOnce('auth_verification', req, res);
+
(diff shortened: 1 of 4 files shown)


middleware/auth.js
```diff
@@ -83,31 +83,26 @@ module.exports = function (params) {
...
-            expectedState = transient.getOnce('state', req, res);
-            const max_age = parseInt(
-              transient.getOnce('max_age', req, res),
-              10
-            );
-            const code_verifier = transient.getOnce('code_verifier', req, res);
-            const nonce = transient.getOnce('nonce', req, res);
+            const authVerification = transient.getOnce('auth_verification', req, res);
+
+            const { max_age, code_verifier, nonce, state } = authVerification ? JSON.parse(authVerification) : {} ;
```
davidpatrick, line 92: We could add a check for `authVerification` and throw an error if the cookie doesn't exist.  Right now it currently falls into `BadRequestError: checks.state argument is missing` https://github.com/auth0/express-openid-connect/issues/145 which isn't very helpful for debugging.

adamjmcgrath: We can look at that in another story

adamjmcgrath (approved):

adamjmcgrath (changes requested):
actually looks like you have some failing tests

adamjmcgrath:
You need to update the callback tests, eg:

```
auth_verification: JSON.stringify({
  nonce: '__test_nonce__',
  state: '__test_state__',
}),
```
For https://github.com/auth0/express-openid-connect/blob/master/test/callback.tests.js#L118-L121

adamjmcgrath (dismissed):

adamjmcgrath (approved):

PR #171, opened 2020-12-16
afterCallback Hook [SDK-1728]

davidpatrick:
This PR introduces an afterCallback to allow the user to cover the following usecases:

1. Validating additional Claims in the ID token
> Given I configure an organisation in Auth0 org_abc123, when I login with the SDK, then I can verify the org_id claim that it equals org_abs123, and if it is successful I can finish the login, and if it fails I can fail the login with a 401

2. Adding additional profile information from /userinfo
> Given my id_token does not contain all my profile claims, when I login, then I can fetch extra claims from /userinfo, and I can add them to the session

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -324,6 +324,25 @@ interface ConfigParams {
   getLoginState?: (req: OpenidRequest, options: LoginOptions) => object;
 
   /**
+   * Function for custom callback handling after receiving tokens and before redirecting
+   * This can be used for handling token storage, making userinfo calls, claim validation, etc.
+   *
+   * ```js
+   * app.use(auth({
+   *   ...
+   *   afterCallback(req, res, session, decodedState) {
+   *     ...
+   *     return {
+   *       ...session,
+   *       ...additionalUserClaims
+   *     };
+   *   }
+   * }))
+   * ``
+   */
+  afterCallback?: (req: OpenidRequest, res: OpenidResponse, session: object, decodedState: object) => object;
+
+  /**
    * Array value of claims to remove from the ID token before storing the cookie session.
    * Default is `['aud', 'iss', 'iat', 'exp', 'nbf', 'nonce', 'azp', 'auth_time', 's_hash', 'at_hash', 'c_hash' ]`
    */
--- a/index.d.ts
+++ b/index.d.ts
@@ -8,6 +8,21 @@ import {
 import { Request, Response, RequestHandler } from 'express';
 
 /**
+ * Session object
+ */
+interface Session {
+  /**
+   * Values stored in an authentication session
+   */
+  id_token: string;
+  access_token: string;
+  refresh_token: string;
+  token_type: string;
+  expires_at: string;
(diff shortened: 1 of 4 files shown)


middleware/auth.js
```diff
@@ -108,15 +109,11 @@ module.exports = function (params) {
...
-          // intentional clone of the properties on tokenSet
-          Object.assign(req[config.session.name], {
-            id_token: tokenSet.id_token,
-            access_token: tokenSet.access_token,
-            refresh_token: tokenSet.refresh_token,
-            token_type: tokenSet.token_type,
-            expires_at: tokenSet.expires_at,
-          });
+          if (config.afterCallback) {
+            session = await config.afterCallback(req, res, session);
```
adamjmcgrath, line 113: You ned to pass in the `decodedState` (or change the ts defs) - although I'm inclined to suggest passing it in since we'll need to do this in the next sdk https://github.com/auth0/nextjs-auth0/blob/master/src/handlers/callback.ts#L57

index.d.ts
```diff
@@ -323,6 +323,25 @@ interface ConfigParams {
...
+   *     ...
+   *     return {
+   *       ...session,
+   *       ...additionalUserClaims
+   *     };
+   *   }
+   * }))
+   * ``
+   */
+  afterCallback?: (req: OpenidRequest, res: OpenidResponse, session: object, decodedState: object) => object;
```
adamjmcgrath, line 343: Let's call it a `tokenSet` and type it as a `TokenSet`

adamjmcgrath (changes requested):
Looks good, couple of comments

adamjmcgrath:
@panva what do you think of the name? (and do you have any other comments)

the [next SDK](https://github.com/auth0/nextjs-auth0/blob/master/src/handlers/callback.ts#L57) calls it `onUserLoaded` - I'll change it to match whatever we decide for this sdk

index.d.ts
panva: We do not expose any internals of the `openid-client` library, i suggest we don't start doing so here.

Nevertheless the types `session: object, decodedState: object` don't allow developers to work with the arguments without explicitly casting them to something they know is on them. I suggest typing these with what we know is there + having a `[propName: string]: any` indexable property.

Furthermore the function type should allow the result to be a promise. 

```
type AsyncOrSync<T> = PromiseLike<T> | T

afterCallback?: (...) => AsyncOrSync<resultInterface>;
```

panva (changes requested):

panva:
> @panva what do you think of the name? (and do you have any other comments)
> 
> the [next SDK](https://github.com/auth0/nextjs-auth0/blob/master/src/handlers/callback.ts?rgh-link-date=2020-12-17T10%3A41%3A01Z#L57) calls it `onUserLoaded` - I'll change it to match whatever we decide for this sdk

I am indifferent as to what to name this.

adamjmcgrath (approved):

Issue #143, opened 2020-09-30
V2 cookies are huge


piec:
Hi,

Thank you for your work on v2.
In my use case I need to call userinfo and the v2 api [makes it really simple](https://github.com/auth0/express-openid-connect/blob/master/EXAMPLES.md#6-calling-userinfo). In v1 I have to keep around users' tokenset with express-session for example.

The issue I have is the size of the appSession cookies (plural because they are chunked in several cookies).
![image](https://user-images.githubusercontent.com/781471/94672769-0308bc00-0316-11eb-8a2d-9792f73300b1.png)
A bit more than 4k in my case, is it due to my setup? (I guess a tokenset is stored in the cookies which explains the simplified userinfo api). V1 appSession cookies are ~500 bytes. 
For each request+response from the browser to the api the cookies are sent as part of the request and updated cookies are sent from the api.
I think I wouldn't have noticed immediately but the default configuration of nginx (used as reverse proxy in front of the api) doesn't let these large headers go through: (due to the cookie size I think)
```
[error] 27#27: *897 upstream sent too big header while reading response header from upstream, client: 172.18.0.1, server: , request: "POST /api/auth/callback HTTP/1.1", upstream: "http://172.18.0.5:3001/api/auth/callback"
```
nginx can be configured to allow this using for example:
```conf
proxy_buffer_size          128k;
proxy_buffers              4 256k;
proxy_busy_buffers_size    256k;
```

What do you think about it?
The tokenset has to be stored somewhere but it's not always needed for very simple usecases right?

adamjmcgrath:
Hi @piec - thanks for raising this

In v2 we wanted to simplify the advice on getting/persisting access tokens. So we went for storing them, along with ID tokens, in the cookie session by default.

The tradeoff we made was that now the cookies can get larger, and so we added cookie chunking in to avoid hitting browser limits on individual cookie size (~4kb)

If I run the sample app with an Auth0 account and set `authorizationParams.response_type` `'code'` (So I can get an AT to access /userinfo), my token set looks like:

```json
{
	"id_token": "[a JWT ~1,900 chars long]",
	"access_token": "An opaque token ~30 chars long",
	"token_type": "Bearer",
	"expires_at": 1601551231
}
```

Which serializes (encrypted) to about 2.8kb. If my access token was a JWT as well, or I had a bunch more claims in my ID token this size would increase, perhaps to slightly over 4kb, so in the case of nginx I would have to update my server configuration to allow for this.

[...]

PR #190, opened 2021-02-09
Custom Session Stores

davidpatrick:
Introduce the ability for a user to specify a custom session store for the session data to be stored.  Custom stores will need to be compatible with express session middleware https://github.com/expressjs/session#compatible-session-stores

See: #143


examples/custom-session-store.js
```diff
@@ -0,0 +1,30 @@
+const express = require('express');
+const { auth } = require('../');
+
+const app = express();
+const IN_MEMORY_SESSION = {};
+
+app.use(
+  auth({
+    idpLogout: true,
+    sessionStore: function Store(config) {
```
adamjmcgrath, line 10: the API should be `sessionStore: { get: ...., set: ... }`

lib/session/middleware.js
```diff
@@ -0,0 +1,118 @@
...
+    attachSessionObject(req, sessionName, session || {});
+
+    // when the request ends, save the data to the session store
+    if (isAsync) {
+      // to handle async saving we have to patch express res.end
+      const { end: _end } = res
+      res.end = function resEnd (...args) {
+        if (!req[sessionName] || !Object.keys(req[sessionName]).length) {
+          debug('session was deleted or is empty, clearing all matching session cookies');
+          cookie.clearSession(req, res);
```
adamjmcgrath, line 65: we need to implement `store.destroy` to avoid memory leaks

adamjmcgrath (approved):
lgtm - see what @panva thinks too

test/appSession.customStore.tests.js
```diff
@@ -0,0 +1,156 @@
...
+const request = require('request-promise-native').defaults({
+  simple: false,
+  resolveWithFullResponse: true,
+});
+
+const appSession = require('../lib/appSession');
+const { get: getConfig } = require('../lib/config');
+const { create: createServer } = require('./fixture/server');
+const redis = require('redis-mock');
+const RedisStore = require('connect-redis')({ Store: class Store {} });
```
adamjmcgrath, line 12: This is bit of a pain, most session stores expect the `express-session` instance to define the store interface

lib/appSession.js
```diff
@@ -229,7 +289,13 @@ module.exports = (config) => {
...
     }
 
-    onHeaders(res, setCookie.bind(undefined, req, res, { iat }));
+    const { end: origEnd } = res;
+    res.end = async function resEnd(...args) {
+      await store.set(existingSessionValue, req, res, {
+        iat,
+      });
+      origEnd.call(res, ...args);
+    };
```
panva, line 298: How are storage errors propagated?

lib/config.js
```diff
@@ -1,9 +1,9 @@
 const Joi = require('@hapi/joi');
-const clone = require('clone');
```
panva, line 2: @adamjmcgrath what is the reason for this clone in the first place? I can't remember.

@davidpatrick what is the reason for this deletion?

lib/appSession.js
adamjmcgrath: I'm assuming you'd log them on the store instance (or the client instance that the store is using, eg https://github.com/tj/connect-redis#how-to-log-redis-errors)

I don't _think_ express-sessions has a better solution than this

panva:
In the example i'd like to see how to use this feature with the existing list of `express-session` stores.

lib/config.js
adamjmcgrath: I deleted it because I didn't want to clone the store instance (in case user's added event listeners to it etc.)

I believe it was there to prevent unwanted mutations to the user's config, but objects passed into joi are effectively immutable and mutating the store instance is desirable (adding event listeners etc.) - so I don't think it's necessary (I removed it for the Next SDK as well)

adamjmcgrath:
> In the example i'd like to see how to use this feature with the existing list of `express-session` stores.

👍 We can change the example to use something like https://www.npmjs.com/package/memorystore? (don't want the user to have to run something like redis or memcached to get an example running)

test/appSession.customStore.tests.js
davidpatrick: This **_is_** a pretty awkward import.  We could alternatively have them pass in the connector (default export) with store options, and then expose then instantiated the store.

lib/appSession.js
adamjmcgrath: Hmmm... actually, maybe we do need to do something with the `next` https://github.com/expressjs/session/blob/master/index.js#L337

I thought `res.end` was too late to call it but maybe not...

adamjmcgrath (changes requested):
I added a failing test for the storage error handling https://github.com/auth0/express-openid-connect/pull/190/commits/19066962bfa9d898a75b8d608a494437f50a910c

adamjmcgrath (approved):

PR #192, opened 2021-02-17
Update README.md

joshcanhelp:
- Fix code sample in the docs
- Link directly to configuration
- Fix eslint picking up on coverage

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -197,15 +197,18 @@ interface LogoutOptions {
 /**
  * Configuration parameters passed to the `auth()` middleware.
  *
- * {@link ConfigParams.issuerBaseURL issuerBaseURL}, {@link ConfigParams.baseURL baseURL}, {@link ConfigParams.clientID clientID}, {@link ConfigParams.clientSecret clientSecret}
- * and {@link ConfigParams.secret secret} are required but can be configured with environmental variables:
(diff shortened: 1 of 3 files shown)


stevehobbsdev (approved):
🎉

joshcanhelp:
Just FYI, look like there is a flaky test for "logout route > should perform a distributed logout" ... docs change caused a fail and re-running fixed it.

index.d.ts
```diff
@@ -197,13 +197,14 @@ interface LogoutOptions {
 /**
  * Configuration parameters passed to the `auth()` middleware.
  *
- * {@link ConfigParams.issuerBaseURL issuerBaseURL}, {@link ConfigParams.baseURL baseURL}, {@link ConfigParams.clientID clientID}
+ * {@link ConfigParams.issuerBaseURL issuerBaseURL}, {@link ConfigParams.baseURL baseURL}, {@link ConfigParams.clientID clientID}, {@link ConfigParams.clientSecret clientSecret}
```
adamjmcgrath, line 200: `clientSecret` isn't required by default, only if you're doing code flow

adamjmcgrath (changes requested):

index.d.ts
joshcanhelp: True 👍 but it can be included with env vars so I'll re-word a bit.

adamjmcgrath (approved):

adamjmcgrath:
Thanks @joshcanhelp!

PR #208, opened 2021-03-30
Update index.d.ts

### LINKED ISSUE #206: Insecure example in the documentation?
Hi all,

I was trying to write some `afterCallback` when I noticed this in the example: https://github.com/auth0/express-openid-connect/blob/master/EXAMPLES.md#8-validate-claims-from-an-id-token-before-logging-a-user-in

It's not clear that `express-openid-connect` *verifies* the request from the IdP in the authentication lifecycle by the point the callback is called.

In the documentation, it is only noted that: "Function for custom callback handling after receiving tokens" - again, because of the wording the user can't be sure that the tokens are to be trusted or not.

If you don't know whether to trust the token or not, then the obvious thing would be to verify the JWT's signature, but on the documentation (see the first link in this issue), it only *decodes* the `id_token`, and not *verify* it.

So, my question is: can we *trust* the request that populated the `session.id_token` **AND** that `express-openid-connect` has verified that trust before `afterCallback` is called? Or can we actually NOT trust aforementioned request and, therefore, the `id_token`?

If it's the former, then it should be noted in the documentation for `afterCallback`. If not, the documentation for usage of `afterCallback` should be updated to use `JWT.verify` instead (so either way the documentation would need to be updated).

Thank you.

adamjmcgrath:
Make it clear in the documentation that the ID Token is validated before the `afterCallback` hook is called

fixes #206

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -343,7 +343,7 @@ interface ConfigParams {
   getLoginState?: (req: OpenidRequest, options: LoginOptions) => object;
 
   /**
-   * Function for custom callback handling after receiving tokens and before redirecting
+   * Function for custom callback handling after receiving and validating the ID Token and before redirecting.
    * This can be used for handling token storage, making userinfo calls, claim validation, etc.
    *
    * ```js


stevehobbsdev (approved):
Looks good, it's just the tests are failing.

PR #228, opened 2021-04-29
Tokenendpoint Parameters [SDK-1730]

davidpatrick:
## Description
Given I configure the SDK
When I add custom parameters to the tokenEndpointParams argument
Then those additional parameters are reflected in the oauth/token request for the code exchange

Given I configure the SDK
When I add custom parameters to the tokenEndpointParams argument
Then those additional parameters are reflected in the oauth/token request for the refresh grant

Given I configure the SDK
When I add custom parameters to the tokenEndpointParams argument
And I call req.oidc.accessToken.refresh({ tokenEndpointParams }) method with extra params
Then both the config and refresh argument params are passed to the token endpoint with the params argument taking precedence

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -453,7 +453,7 @@ interface ConfigParams {
   clientAuthMethod?: string;
 
   /**
-   * Additional request body properties to be sent to the `oauth/token` endpoint during code exchange or token refresh.
+   * Additional request body properties to be sent to the `token_endpoint` during authorization code exchange or token refresh.
    */
   tokenParameters?: {
     [key: string]: unknown;
--- a/index.d.ts
+++ b/index.d.ts
@@ -451,6 +451,13 @@ interface ConfigParams {
    * String value for the client's authentication method. Default is `none` when using response_type='id_token', otherwise `client_secret_basic`.
    */
   clientAuthMethod?: string;
+
+  /**
+   * Additional extra request body properties to be sent to the AS during code or refresh token exchange
+   */
+  tokenParameters?: {
+    [key: string]: unknown;
+  }
 }
 
 interface SessionStorePayload {
--- a/index.d.ts
+++ b/index.d.ts
@@ -453,7 +453,7 @@ interface ConfigParams {
   clientAuthMethod?: string;
 
   /**
-   * Additional extra request body properties to be sent to the AS during code or refresh token exchange
+   * Additional request body properties to be sent to the `oauth/token` endpoint during code exchange or token refresh.
    */
   tokenParameters?: {
     [key: string]: unknown;
--- a/index.d.ts
+++ b/index.d.ts
@@ -455,7 +455,7 @@ interface ConfigParams {
   /**
    * Additional request body properties to be sent to the `token_endpoint` during authorization code exchange or token refresh.
    */
-  tokenParameters?: TokenParameters
+  tokenEndpointParams?: TokenParameters
 }
 
 interface SessionStorePayload {
--- a/index.d.ts
+++ b/index.d.ts
@@ -455,9 +455,7 @@ interface ConfigParams {
   /**
    * Additional request body properties to be sent to the `token_endpoint` during authorization code exchange or token refresh.
    */
-  tokenParameters?: {
-    [key: string]: unknown;
-  }
+  tokenParameters?: TokenParameters
 }
 
 interface SessionStorePayload {
@@ -630,7 +628,11 @@ interface AccessToken {
    * }
    * ```
    */
-  refresh(): Promise<AccessToken>;
+  refresh(params?: TokenParameters): Promise<AccessToken>;
+}
+
+interface TokenParameters {
+  [key: string]: unknown;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


panva (commented):
Should we allow this to be specified dynamically with the request's context, e.g. with a function instead of a fixed set of properties?

index.d.ts
```diff
@@ -451,6 +451,13 @@ interface ConfigParams {
    * String value for the client's authentication method. Default is `none` when using response_type='id_token', otherwise `client_secret_basic`.
    */
   clientAuthMethod?: string;
+
+  /**
+   * Additional extra request body properties to be sent to the AS during code or refresh token exchange
+   */
+  tokenParameters?: {
```
adamjmcgrath, line 458: `tokenEndpointParams`

You also need to update the signature for `refresh` https://github.com/auth0/express-openid-connect/blob/b73edb95455ad6a04b51290f18eb71db54d068cd/index.d.ts#L633

index.d.ts
```diff
@@ -451,6 +451,13 @@ interface ConfigParams {
    * String value for the client's authentication method. Default is `none` when using response_type='id_token', otherwise `client_secret_basic`.
    */
   clientAuthMethod?: string;
+
+  /**
+   * Additional extra request body properties to be sent to the AS during code or refresh token exchange
```
adamjmcgrath, line 456:
```suggestion
   * Additional request body properties to be sent to the `oauth/token` endpoint during code exchange or token refresh.
```

lib/context.js
```diff
@@ -15,11 +15,17 @@ function isExpired() {
   return tokenSet.call(this).expired();
 }
 
-async function refresh() {
+async function refresh(params) {
```
adamjmcgrath, line 18: `refresh({ tokenEndpointParams })` Let's make this a config object and also call it `tokenEndpointParams`

test/callback.tests.js
```diff
@@ -583,6 +583,93 @@ describe('callback response_mode: form_post', () => {
...
+    });
+
+    const authOpts = {
+      ...defaultConfig,
+      clientSecret: '__test_client_secret__',
+      authorizationParams: {
+        response_type: 'code id_token',
+        audience: 'https://api.example.com/',
+        scope: 'openid profile email read:reports offline_access',
+        tokenEndpointParams: {
```
adamjmcgrath, line 598: This shouldn't be nested in `authorizationParams` - which is why `longeLiveToken` isn't being passed to the token endpoint.

test/callback.tests.js
```diff
@@ -583,6 +583,93 @@ describe('callback response_mode: form_post', () => {
...
+    const router = auth(authOpts);
+    router.get('/refresh', async (req, res) => {
+      const accessToken = await req.oidc.accessToken.refresh({ force: true });
+      res.json({
+        accessToken,
+        refreshToken: req.oidc.refreshToken,
+      });
+    });
+
+    const { tokens, jar } = await setup({
```
adamjmcgrath, line 612: You can also get the `tokenReqBody` for the code exchange, then assert the `tokenEndpointParams` have been added there too.

```js
const { tokens, jar, tokenReqBody } = await setup({ ... });

assert.match(tokenReqBody, /longeLiveToken=true/);
```

adamjmcgrath (changes requested):

adamjmcgrath:
@panva 

> Should we allow this to be specified dynamically with the request's context, e.g. with a function instead of a fixed set of properties?

Good point, I forgot about this

We don't do it with `authorizationParams` because you can customise the login handler and pass `authorizationParams` in depending on the request context.

```js
app.get('/login', (req, res) => {
  res.oidc.login({ authorizationParams: { ... something related to req ... } });
})
```

I'd quite like to resolve this when it comes up with `callback` and `tokenEndpointParams` as well, with the caveat that it's harder to overdide the callback

```js
// or `get`
app.post('/callback', express.urlencoded({ extended: false }), (req, res) => {
  res.oidc.callback({ tokenEndpointParams: { ... somthing related to req ... } });
});
```

index.d.ts
```diff
@@ -453,7 +453,7 @@ interface ConfigParams {
   clientAuthMethod?: string;
 
   /**
-   * Additional extra request body properties to be sent to the AS during code or refresh token exchange
+   * Additional request body properties to be sent to the `oauth/token` endpoint during code exchange or token refresh.
```
panva, line 456:
```suggestion
   * Additional request body properties to be sent to the `token_endpoint` during authorization code exchange or token refresh.
```

index.d.ts
```diff
@@ -451,6 +451,11 @@ interface ConfigParams {
    * String value for the client's authentication method. Default is `none` when using response_type='id_token', otherwise `client_secret_basic`.
    */
   clientAuthMethod?: string;
+
+  /**
+   * Additional request body properties to be sent to the `token_endpoint` during authorization code exchange or token refresh.
+   */
+  tokenParameters?: TokenParameters
```
adamjmcgrath, line 458: `tokenEndpointParams`

adamjmcgrath (changes requested):
just need to fix the type defs for `tokenEndpointParams`

adamjmcgrath (approved):

PR #230, opened 2021-05-05
[SDK-2533] Swallow error on silent auth

### LINKED ISSUE #219: login_required exeption on silent login
### Describe the problem

After fixing Silent Login Bug (#212 #217) it seems to get a step forward.
But when i have no valid session the callback from the identity called in this way -> 
`... /signin-oidc?error=login_required&state=eyJ ...`

And this produce this exeption -> 
```
BadRequestError: login_required
    at C:\Projekte\auth-test-app-1\node_modules\express-openid-connect\middleware\auth.js:109:31
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

I think this exeption should not thrown when silent login happens. But is it a problem in lib or in my config?

Demonstration application -> [Auth Test App](https://github.com/dkleber89/auth-test-app-1)

No exeption on silent login when no valid session is available.

### Environment

- Lib Version v2.3.1
- Express Lib 4.17.0
- Identity Server 4

adamjmcgrath:
### Description

When the application attempts silent auth, callback errors should be swallowed.

### References

fixes #219 

### Testing

### Checklist

### DIFF
--- a/end-to-end/attempt-silent-login.test.js
+++ b/end-to-end/attempt-silent-login.test.js
@@ -1,0 +1,78 @@
+const { assert } = require('chai');
+const puppeteer = require('puppeteer');
+const provider = require('./fixture/oidc-provider');
+const {
+  baseUrl,
+  start,
+  login,
+  runExample,
+  stubEnv,
+  goto,
+} = require('./fixture/helpers');
+
+describe('attempt silent login', async () => {
+  let authServer;
+  let appServer;
+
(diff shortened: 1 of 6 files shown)


davidpatrick (commented):
Looks good.  We should probably add the new example to https://github.com/auth0/express-openid-connect/blob/master/EXAMPLES.md

davidpatrick (approved):

PR #237, opened 2021-05-20
Chunked cookies should not exceed browser max

davidpatrick:
We currently allow 96 characters leeway for cookie attribute size ([browser cookie max size (4096 Bytes)](https://tools.ietf.org/html/rfc6265#section-6.1) minus [CHUNK_BYTE_SIZE](https://github.com/auth0/express-openid-connect/blob/master/lib/appSession.js#L17))

This is not enough for scenarios where the user provides many cookie options or longer cookie domain or path options.

Since the cookie attributes are dynamic and can be quite long, we can calculate the cookie attributes length from serializing an empty cookie with the same options and measuring it.

Same strategy as done over in https://github.com/auth0/nextjs-auth0/pull/301

### DIFF
--- a/lib/appSession.js
+++ b/lib/appSession.js
@@ -49,6 +49,12 @@ module.exports = (config) => {
     rollingDuration,
   } = config.session;
 
+  const { transient: emptyTransient , ...emptyCookieOptions } = cookieConfig;
+  emptyCookieOptions.expires = emptyTransient ? 0 : new Date();
+
+  const emptyCookie = cookie.serialize(`${sessionName}.0`, '', emptyCookieOptions);
+  const cookieChunkSize = MAX_COOKIE_SIZE - emptyCookie.length;
+
   let keystore = new JWKS.KeyStore();
 
   secrets.forEach((secretString, i) => {
@@ -91,11 +97,8 @@ module.exports = (config) => {
     { uat = epoch(), iat = uat, exp = calculateExp(iat, uat) }
   ) {
     const cookies = req[COOKIES];
-    const cookieOptions = {
-      ...cookieConfig,
-      expires: cookieConfig.transient ? 0 : new Date(exp * 1000),
-    };
-    delete cookieOptions.transient;
+    const { transient: cookieTransient , ...cookieOptions } = cookieConfig;
+    cookieOptions.expires = cookieTransient ? 0 : new Date(exp * 1000);
 
     // session was deleted or is empty, this matches all session cookies (chunked or unchunked)
     // and clears them, essentially cleaning up what we've set in the past that is now trash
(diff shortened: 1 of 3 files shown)


lib/appSession.js
```diff
@@ -115,25 +116,46 @@ module.exports = (config) => {
         'found session, creating signed session cookie(s) with name %o(.i)',
         sessionName
       );
+      const emptyCookie = cookie.serialize(`${sessionName}.0`, '', cookieOptions);
```
adamjmcgrath, line 119: Better to do this outside of a request, then it only happens once at boot time

adamjmcgrath (changes requested):
lgtm - just one suggested change

adamjmcgrath:
@davidpatrick - the tests are failing because the max header size for Node 12 is 8KB

You could up it to 16kb on the circle CI environment `NODE_OPTIONS=--max-http-header-size=16384` https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-container

Or rewrite the tests to keep the header <8KB

adamjmcgrath (approved):

lib/appSession.js
```diff
@@ -90,19 +96,17 @@ module.exports = (config) => {
...
     { uat = epoch(), iat = uat, exp = calculateExp(iat, uat) }
   ) {
-    const cookieOptions = {
-      ...cookieConfig,
-      expires: cookieConfig.transient ? 0 : new Date(exp * 1000),
-    };
-    delete cookieOptions.transient;
+    const cookies = req[COOKIES];
+    const { transient: cookieTransient , ...cookieOptions } = cookieConfig;
+    cookieOptions.expires = cookieTransient ? 0 : new Date(exp * 1000);
```
adamjmcgrath, line 101: @davidpatrick I know this was the existing logic, but I noticed it's not correct.

If you want a session cookie `expires` should not be defined (it shouldn't be set to 0)

> If (expires is) unspecified, the cookie becomes a session cookie. A session finishes when the client shuts down, and session cookies will be removed.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes

See nextjs-auth0 for reference https://github.com/auth0/nextjs-auth0/blob/main/src/auth0-session/cookie-store.ts#L44-L46 and https://github.com/auth0/nextjs-auth0/blob/main/tests/auth0-session/cookie-store.test.ts#L262-L271

Would you mind fixing it and adding a regression test? Either in this PR or another one before you do the release

adamjmcgrath: @davidpatrick - actually, aplogies, ignore me - I've just tested it and I get a sessoion cookie fine - so don't worry

PR #251, opened 2021-06-25
feat(patch): Added httpTimeout Option

### LINKED ISSUE #249: allow configuration of oidc client timeout
### Describe the problem you'd like to have solved

We have an internal oidc issuer whose response time is less than performant.  To accommodate them, I'd like to be able to configure the http timeout.   If I leverage the `openid-client` package directly without using the sdk, I can do the following:

```
const { custom } = require('openid-client')

custom.setHttpOptionsDefaults({
  timeout: 10000, // set custom timeout
})
```

This doesn't work with the SDK, however, because the timeout is always overridden to 5000ms in [client.js](https://github.com/auth0/express-openid-connect/blob/master/lib/client.js#L35). 

### Describe the ideal solution

Depends.  If you want to control the interface, you could expose a `timeout` option on the `ConfigParams` object.  Otherwise the quick solution would be to just to check if the timeout if null / undefined just before you set it in  [client.js](https://github.com/auth0/express-openid-connect/blob/master/lib/client.js#L35), and let consumers override default http options using the underlying open-id client package as shown above.

## Alternatives and current work-arounds

Haven't considered alternatives.  If the above sounds like a good idea let me know and I'll submit a PR.

Thanks!

jmacvey:

### Description

Facilitates `httpTimeout` configuration option (in milliseconds) on `ConfigParams` interface.  This allows package consumers to configure the http timeout for requests against less-than-performant issuers.

### References

Closes #249.

### Testing

**Language:** Javascript
**Node Versions Tested:** 12.19.0 

- [X] This change adds test coverage for new/changed/fixed functionality

### Checklist

- [X] All active GitHub checks for tests, formatting, and security are passing

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -463,7 +463,7 @@ interface ConfigParams {
   tokenEndpointParams?: TokenParameters;
 
   /**
-   * Default http timeout for oidc client requests in milliseconds.  Default is `5000`.
+   * Http timeout for oidc client requests in milliseconds.  Default is 5000.
    */
   httpTimeout?: number;
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -463,7 +463,7 @@ interface ConfigParams {
   tokenEndpointParams?: TokenParameters;
 
   /**
-   * Http timeout for oidc client requests in milliseconds.  Default is 5000.
+   * Http timeout for oidc client requests in milliseconds.  Default is 5000.   Minimum is 500.
    */
   httpTimeout?: number;
 }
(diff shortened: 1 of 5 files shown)


jmacvey:
Note: I have not updated documentation explicitly in `.html` files under docs folder.  I'm assuming there's a tool that does that, but if not I'll go and update them line by line.  Let me know.

adamjmcgrath (approved):

adamjmcgrath:
Thanks @jmacvey!

Issue #200, opened 2021-03-08
Multiple baseUrls for the same application?


adbutterfield:
### Describe the problem you'd like to have solved

Not sure if this is a feature request, or we're just overlooking something and need someone to point us in the right direction, but here it goes...

The environments we deploy our applications on can be accessed my multiple URLs (usually one internet and one intranet). However it looks like this package (or maybe even Auth0 in general?) can't handle this setup, because we have to pass `baseUrl` to the middleware. But technically, the app instance has more than one `baseUrl`. 

### Describe the ideal solution

Do we need to explicitly pass the `baseUrl`? Couldn't this be inferred from the request object?

I noticed in [context.js](https://github.com/auth0/express-openid-connect/blob/3adcd922da3a1b4b4d516836d14e31af234706a3/lib/context.js#L181) it does default to using `req.originalUrl` for the `returnUrl`, and as far as I could tell, that's the only place where `baseUrl` is used, for the `returnUrl`. 

Of course I only glanced through the source code, so I'm not 100% sure how everything works.

## Alternatives and current work-arounds

~~I have not yet been able to think of any workarounds.~~

Update: see comment below.

### Additional information, if any

I also did open a topic on the [Auth0 Community](https://community.auth0.com/t/multiple-domains-for-the-same-client/59053), because we hoping to get an answer if there's something we're overlooking as soon as possible. Or else we'll have to look at using a different middleware (which would make me sad, because this package is pretty awesome).

adbutterfield:
Figured out a potential workaround. Setting the baseUrl dynamically like so:

```javascript
import { auth } from 'express-openid-connect';

const configParams = {
 // ...my config
};

const auth0Mw = (req, res, next) => {
  const baseURL = `${req.protocol}://${req.header('x-forwarded-host') || req.get('host')}`;
  return auth({ ...configParams, baseURL })(req, res, next);
};
```

I found that `req.originalUrl` won't work for us, because of reverse proxy. So I'm trying to get the host from the headers instead (unfortunately the headers aren't configured correctly on the environment, but we're working on that). But I guess this is why you explicitly need to pass the `baseUrl` then.

I guess my question is then, do you see any issue with this approach?

adamjmcgrath:
Hi @adbutterfield - thanks for raising this

Your suggestion looks fine, there's a little bit of overhead to calling `auth()` - so I'd try and do that when the app mounts rather than for each request, eg if you have 2 base urls for internet and intranet

```js
[...]

PR #289, opened 2021-10-15
Baseurl as func

tommyk:
# overview

this change was needed so we could change the baseUrl based on the incoming request, not just a static url.  i just wanted to get this PR going and get feedback as to what would be needed to get this mainlined.

would appreciate any feedback and/or directions on making this mergeable, thanks!

this is an example of how we are currently using it in case that helps.
```js
baseURL: (context: any) => {
    return `https://${context.req.get('host')}`;
  },
```

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -168,7 +168,7 @@ class ResponseContext {
 
     const baseUrl =
       typeof config.baseURL === 'function'
-        ? config.baseURL(this)
+        ? config.baseURL(weakRef(this))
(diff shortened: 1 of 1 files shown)


adamjmcgrath:
Thanks for the feedback @tommyk - see [my comment](https://github.com/auth0/express-openid-connect/issues/200#issuecomment-793672607) about adding the `redirectUri` option to the callback handler for how this might be resolved in the future

Issue #162, opened 2020-11-20
Support JWT-Based Client Authentication


madaster97:
### Describe the problem you'd like to have solved
I am trying to develop against on OIDC provider that only accepts private_key_jwt client authentication, and am unable to with this middleware.

### Describe the ideal solution
Add support for private_key_jwt and client_secret_jwt client authentication methods. In the startup config, a field for a JWKs set would be best, where the specific KID to be used on a request could be routed on login (similar to solution in this <a href="https://github.com/auth0/express-openid-connect/issues/161">issue</a>). If the JWK only had one key, then that could be used as a default client authentication value.

## Alternatives and current work-arounds
None for this library. I'm currently trying to wrap express around openid-client myself, but that loses a lot of the awesome functionality in this lib :)

adamjmcgrath:
Hey @madaster97 - thanks for your suggestion, let me kick the idea round with my team and get back to you on it. 

Response may be a bit slow next week because of thanksgiving

madaster97:
Thanks Adam. Have a nice Thanksgiving!

madaster97:
Adding one more item here. Some authorization servers support "none" for client authentication.

For issuers that support public clients, it would be great if this library could be used for authorization code or hybrid flows without using client authentication.

adamjmcgrath:
Hey @madaster97 

Regarding `private_key_jwt` and `client_secret_jwt` - adding these new client methods is out of scope for us, mainly because Auth0 doesn't support them. We'd of course be happy to look at a pull request - with the caveat that we may be opinionated about how it's implemented 😄 

We support the 'none' `clientAuthMethod` with implicit, but we won't support it with hybrid since this is a backend SDK that can keep a secret

madaster97:
Thanks Adam. I'll consider it! Right now I'm just using openid-client manually with express. Also, the none bit makes sense. Thanks!

PR #291, opened 2021-10-24
Jwt client authn

madaster97:
### Description
This PR adds JWT based client authentication to the library (building on Adam's initial implementation). All signing algorithms supported by [jose@2.0.5](https://github.com/panva/jose/tree/v2.x#detailed-support-matrix) will be supported here (though upgrading jose versions is an option). This PR also adds a configurable JWKS endpoint, so that the public keys associated with the client can be distributed.

### References
Following up on my [issue](https://github.com/auth0/express-openid-connect/issues/162) for this enhancement.
GitHub [Project](https://github.com/madaster97/express-openid-connect/projects/1) where I documented the additions

### Testing
There's an example JWT client that @adamjmcgrath already added. I've been using that to test, but I've been having issues getting it to work against the fixture provider. I think the `run_example.js` script is accidentally setting up a client_secret client instead of a jwt one. I've been testing against an example Openid Provider that supports private_key_jwt clients and it's worked so far.

#### TBD on testing
- [] Adding testing for the JWKS endpoint

### Checklist

### DIFF
--- a/end-to-end/fixture/oidc-provider.js
+++ b/end-to-end/fixture/oidc-provider.js
@@ -17,6 +17,7 @@ const config = {
   clients: [
     client,
     Object.assign({}, client, {
+      client_secret: undefined,
       client_id: 'jwtca-client',
       token_endpoint_auth_method: 'private_key_jwt',
       jwks: {
--- a/examples/jwt-client-authn.js
+++ b/examples/jwt-client-authn.js
@@ -13,9 +13,12 @@ app.use(
       response_type: 'code',
     },
     clientAuthMethod: 'private_key_jwt',
-    clientAssertionSigningKey: fs.readFileSync(
+    clientAssertionConfig: {
+      signingKey: fs.readFileSync(
       path.join(__dirname, 'private-key.pem')
     ),
+      signingAlg: 'RS256'
+    }
   })
 );
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -86,6 +86,38 @@ async function get(config) {
     );
   }
 
+  const configTokenAuthMethod = config.clientAuthMethod;
+  const issuerTokenAuthMethods = Array.isArray(
+    issuer.token_endpoint_auth_methods_supported
+  )
+    ? issuer.token_endpoint_auth_methods_supported
+    : [];
+  if (!issuerTokenAuthMethods.includes(configTokenAuthMethod)) {
+    debug(
+      'Client authn method %o is not supported by the issuer. ' +
+        'Supported authn methods are %o.',
+      configTokenAuthMethod,
+      issuerTokenAuthMethods
+    );
+  }
+
+  if (config.clientAuthMethod == 'private_key_jwt') {
+    const configTokenAuthAlg = config.clientAssertionConfig.signingKey;
+    const issuerTokenAuthAlgs = Array.isArray(
+      issuer.token_endpoint_auth_signing_alg_values_supported
+    )
+      ? issuer.token_endpoint_auth_signing_alg_values_supported
+      : [];
+    if (!issuerTokenAuthAlgs.includes(configTokenAuthAlg)) {
+      debug(
+        'Client authn signing alg %o is not supported by the issuer. ' +
+          'Supported authn algs are %o.',
+        configTokenAuthAlg,
+        issuerTokenAuthAlgs
+      );
+    }
+  }
+
   const jwks =
     config.clientAuthMethod == 'private_key_jwt'
       ? {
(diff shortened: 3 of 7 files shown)


adamjmcgrath:
Hi @madaster97 - sorry, I haven't forgotten this. It might be a little while before I get round to reviewing it

adamjmcgrath:
Hi @madaster97 - we do have this in the Roadmap to look at at some point over the next few months, so will reopen this when we do

examples/jwt-client-authn.js
```diff
@@ -13,9 +13,12 @@ app.use(
       response_type: 'code',
     },
     clientAuthMethod: 'private_key_jwt',
-    clientAssertionSigningKey: fs.readFileSync(
+    clientAssertionConfig: {
```
adamjmcgrath, line 16: We don't need to specify the alg if we're not publishing a jwks uri

end-to-end/fixture/oidc-provider.js
```diff
@@ -17,6 +17,7 @@ const config = {
   clients: [
     client,
     Object.assign({}, client, {
+      client_secret: undefined,
```
adamjmcgrath, line 20: This shouldn't be necessary

lib/client.js
```diff
@@ -86,17 +86,69 @@ async function get(config) {
...
+    debug(
+      'Client authn method %o is not supported by the issuer. ' +
+        'Supported authn methods are %o.',
+      configTokenAuthMethod,
+      issuerTokenAuthMethods
+    );
+  }
+
+  if (config.clientAuthMethod == 'private_key_jwt') {
+    const configTokenAuthAlg = config.clientAssertionConfig.signingKey;
```
adamjmcgrath, line 105: This would be `signingAlg` - but I would rather avoid requiring the user to specify this if we can

madaster97:
Thanks @adamjmcgrath - Yeah I'm looking into these comments. Right now just trying to get re-setup for testing

examples/jwt-client-authn.js
madaster97: @adamjmcgrath - There are some cases where you need to explicitly provide an alg. For instance you can use an RSA private key with any of RS 256/384/512, and somewhere along the way openid-client or jose are defaulting to RS256 for an RSA private key.

I can think of two options to let callers specify their alg:
1. Keep the current setup, noting that we're passing the alg to openid-client [here](https://github.com/auth0/express-openid-connect/pull/291/files#diff-ca407d959d33ce92a7f7efee354a5ea5e6e61fcd797d32c04205e429a074ed54:~:text=%3F%20%7B-,keys%3A%20%5B,%5D%2C,-%7D) after parsing with jose's JWK module
2. Explicitly require a jwk and inspect the alg (maybe warning if it's missing and that we'll choose a default)

Note that openid-clilent won't throw an error for invalid algs until runtime. For example, if I try to specify ES384 but give an RSA private key I get the following, but not until I actually attempt a grant:
```
BadRequestError: no key found in client jwks to sign a client assertion with using alg ES384
    at C:\Users\adams\source\node\express-openid-connect\middleware\auth.js:138:31
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

What do you think?

adamjmcgrath: Hey @madaster97 - sorry for the delay responding, I was away last week

Your right, we should allow specifying the algorithm, but it should be optional

By default `openid-client` uses one of the `token_endpoint_auth_signing_alg_values_supported` if you don't specify - so I just need to figure out whether the default should be `RS256` or `undefined` (let openid-client pick one from `token_endpoint_auth_signing_alg_values_supported`)

Also, can we make the config flat,eg `clientAssertionSigningKey` and `clientAssertionSigningKeyAlg` (like `idTokenSigningAlg`)

lib/client.js
```diff
@@ -86,17 +86,69 @@ async function get(config) {
...
+      case 'none':
+        return {};
+      case 'private_key_jwt':
+        return {
+          token_endpoint_auth_signing_alg:
+            config.clientAssertionConfig.signingAlg,
+        };
+      case 'client_secret_basic':
+      case 'client_secret_post':
+        return { client_secret: config.clientSecret };
```
adamjmcgrath, line 143: Also , I think there's a use case for a client secret even when using `private_key_jwt` (when the ID Token signing alg is `HS256`) but I will confirm

adamjmcgrath:
> toJWK(false)

Ah yep, gotcha, thanks

Apologies for taking ages reviewing then being in a rush to merge, just is just because of how our work is prioritised behind the scenes.

I'll add you to the PR when I raise the final one against master in case you want to take a look

PR #317, opened 2022-01-26
Add cross domain iframe support for modern browsers

### LINKED ISSUE #315: SameSite cookie config not respected by skipSilentLogin cookie
### Describe the problem

An application embedded within an iframe gets stuck in a redirect loop when attempting to do silent log in and user is not logged in at IDP. 

The issue seems to be related to the fact that the `skipSilentLogin` cookie is [missing a same site attribute](https://github.com/auth0/express-openid-connect/blob/v2.5.2/middleware/attemptSilentLogin.js#L15) and thus is defaulted to Lax by the browser. As the application is embedded within an iframe this cookie will be blocked and not sent to the application backend causing another silent log in attempt. 

Silent log in should be attempted once and not on subsequent page requests.

I was expecting the `skipSilentLogin` cookie to honor the sameSite attribute from `config.session.cookie.sameSite`. That is unless there is a good reason not to, if so help me understand why. 

I've forked the repository myself to see if changes to pickup the sameSite config for `skipSilentLogin` will solve this issue and it seems to be the case. 

### Reproduction

A reproduction would look something like this

1. Have an application embedded within an iframe, in our case this is an app embedded within Teams. 
2. Make sure the user is not logged in at the IDP (as this will trigger additional attempts to log in)
3. Access the application in a way that would trigger a silent log in and where it should end up on the same path after log in.
4. Observe the redirects from the page in 3) -> auth endpoint (for silent log in) -> callback with error (as user is not logged in)
5. A too many redirects error will be shown in the browser.

### Environment

- **Version of this library used:** 2.5.2
- **Any other relevant information you think would be useful:** Issue detected in Chrome Version 97.0.4692.71 but I suspect all Chrome versions where SameSite is defaulted to Lax will result in this behaviour.

adamjmcgrath:
### Description

Fixing crossdomain iframe login/logout and attemptSilentAuth support. This can be enabled by using `session.cookie.sameSite = 'None'` and `session.cookie.secure = true` (you should run localhost on https for this see: https://auth0.com/docs/libraries/secure-local-development)

### References

fixes #315 

### Testing

To test login/logout from a cross domain iframe:
1. Run an https reverse proxy like caddy, eg `caddy reverse-proxy --from localhost:443 --to localhost:3000`
2. Run the iframe testing example with `baseUrl` set to `https://localhost`
3. Visit `http://localhost:3000/iframe` (running same app on a different domain to the iframe, which is on `https://localhost`)
4. Login and verify that the `auth_verification` transaction cookies have been removed
5. Logout and verify that the `appSession` cookie has been removed

### Checklist


lib/appSession.js
```diff
@@ -213,7 +214,8 @@ module.exports = (config) => {
     ) {
       const hasPrevSession = !!req[COOKIES][sessionName];
       const replacingPrevSession = !!req[REGENERATED_SESSION_ID];
-      const hasCurrentSession = req[sessionName] && Object.keys(req[sessionName]).length;
+      const hasCurrentSession =
```
adamjmcgrath, line 217: These whitespace changes are form a previous `prettier` update

stevehobbsdev (approved):

Issue #322, opened 2022-02-03
Cookies Samesite attribute handling


surensinghkhatana:
### Describe the problem
I have a SPA application that talks to the backend BFF layer (which handles all OAuth flows using express-openid-connect lib). I would like to set `samesite =strict` for the appSession cookie containing the tokens so that the cookie is protected and only sent to my BFF component.

In order to achieve this , I have set the `cookieconfigparams` as `cookie: { httpOnly: true, sameSite: 'Strict', secure: true }` but the problem is that this configuration also sets the  `samesite =strict` for `auth_verification` cookie as well. This prevents the `auth_verification` cookie to be sent to the /callback since it is redirected from a third party auth server domain and that causes the following error : 

`BadRequestError: checks.state argument is missing
    at /home/ec2-user/bff-code/bff-layer/node_modules/express-openid-connect/middleware/auth.js:121:31
    at processTicksAndRejections (node:internal/process/task_queues:96:5)`

I have following questions : 
1) How to set the  `samesite =strict` for the encrypted cookie but keep `samesite =lax` for the  `auth_verification` cookie ? 
2) How to set `__Host-` for the `appSession ` cookie using `cookieconfigparams` ? 
3) Is there any support available for CSRF protection of `appSession ` cookie containing the tokens ?

BitPatty:
Regarding 1) Experiencing the same issue, looks like the configuration for the cookie is being shared and there's no builtin way to override the behaviour:

https://github.com/auth0/express-openid-connect/blob/b850bfd410b78c9c7c0227de9b91c8025d91c597/lib/context.js#L258-L264

adamjmcgrath:
Hi @surensinghkhatana - thanks for raising this

> How to set the samesite =strict for the encrypted cookie but keep samesite =lax for the auth_verification cookie ?

Currently the cookie config is shared for the session and transaction cookies. But your use case sounds like a reasonable one, so I'll keep this open as a possible enhancement.

> How to set __Host- for the appSession cookie using cookieconfigparams ?

You can configure the cookie name using the `session.name` config

> Is there any support available for CSRF protection of appSession cookie containing the tokens ?

The library doesn't provide csrf protection (other than the default cookie config SameSite=Lax). This is a separate concern from login and you should use one of the available express csrf protection middlewares if you want to add this.

surensinghkhatana:
Thanks @adamjmcgrath for quick response. Appreciate your help.

> You can configure the cookie name using the session.name config
[...]

PR #323, opened 2022-02-04
allow configuration of same site attribute on auth_verification cookie

BitPatty:
### Description

This is a breaking change. Allows to use a custom same site configuration for the `auth_verification` cookie. I only included the override of the samesite but it could be extended later on.

### References

#322 

### Testing

### Checklist

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -45,6 +45,14 @@ const paramsSchema = Joi.object({
       .maxArity(1)
       .optional()
       .default(() => defaultSessionIdGenerator),
+    verificationCookie: Joi.object({
+      sameSite: Joi.string()
+        .valid('Lax', 'Strict', 'None')
+        .optional()
+        .default('Lax'),
+    })
+      .default()
+      .unknown(false),
     cookie: Joi.object({
       domain: Joi.string().optional(),
       transient: Joi.boolean().optional().default(false),
--- a/index.d.ts
+++ b/index.d.ts
@@ -571,6 +571,11 @@ interface SessionConfigParams {
   absoluteDuration?: boolean | number;
 
   /**
+   * Configuration parameters used for the verification cookie.
+   */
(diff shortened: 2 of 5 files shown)


lib/config.js
```diff
@@ -45,6 +45,14 @@ const paramsSchema = Joi.object({
       .maxArity(1)
       .optional()
       .default(() => defaultSessionIdGenerator),
+    verificationCookie: Joi.object({
+      sameSite: Joi.string()
+        .valid('Lax', 'Strict', 'None')
+        .optional()
+        .default('Lax'),
```
adamjmcgrath, line 52: `.default(Joi.ref('..cookie.sameSite'))` (or `Joi.ref('session.cookie.sameSite')` if it's in the root of the config)

Then it doesn't have to be a breaking change

lib/config.js
```diff
@@ -45,6 +45,14 @@ const paramsSchema = Joi.object({
       .maxArity(1)
       .optional()
       .default(() => defaultSessionIdGenerator),
+    verificationCookie: Joi.object({
```
adamjmcgrath, line 48: Since the transaction cookie isn't strictly part of the session, can we move this config up to the root.

Also, I'm going to need to think about if other properties need to be configurable.

Also, can we call it `transactionCookie`

adamjmcgrath (changes requested):
Thanks for raising this @BitPatty

lib/config.js
BitPatty: Thanks, didn't know about `ref`.

lib/config.js
BitPatty: Renamed it. I kept it in object format to allow for more options, but I'm unsure if there's currently a use case for modifying the other cookie settings.

adamjmcgrath (approved):

PR #373, opened 2022-06-30
Replace experimental express-oauth2-bearer 

adamjmcgrath:
Replace `express-oauth2-bearer` with `express-oauth2-jwt-bearer` in end to end tests


evansims (approved):
Thanks, @adamjmcgrath! 👍

Issue #397, opened 2022-10-16
when accessing `req.oidc.user`, TypeError: Cannot read properties of undefined  


b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
### Describe the problem

My express server provide public APIs and private APIs both. Private APIs that can be used without auth0 authentication and a private api that requires authentication. example code:

```js
const app = express()

// public api
app.post('/public/api/a', handleA)
app.post('/public/api/b', handleB)
...

// auth0
app.use(auth({...}))

// private api
app.post('/api/c', handleC)
app.post('/api/d', handleD)
...
```

However, in this repo's TypeScript file `index.d.ts`, Adding `oidc` attribute as a non-optional Request and Response like below:

```ts
declare global {
  namespace Express {
    interface Request {
      oidc: RequestContext;
    }

    interface Response {
      oidc: ResponseContext;
    }
  }
}
``` 

So, in the public API handler, `handleA` function, the code below does not occur an error at compile time (tsc), but at runtime, `Cannot read properties of undefined (reading 'user')` this TypeError occurs.

```js
export const handleA = (req, res) => {
  const posts = getUserPosts(req.oidc.user)
}
```

### What was the expected behavior?

Accessing `req.oidc.something` without a type guard should occur `Object is possibly 'undefined'` tsc error.

![image](https://user-images.githubusercontent.com/15520015/196026319-995c49f5-4a07-4770-a15e-3759c3654eb3.png)

### Environment
- express-openid-connect: 2.8.0

adamjmcgrath:
Hi @b6pzeusbc54tvhw5jgpyw8pwz2x6gs - thanks for raising this

> So, in the public API handler, handleA function, the code below does not occur an error at compile time (tsc), but at runtime, Cannot read properties of undefined (reading 'user') this TypeError occurs.

There's a tradeoff here, the SDK assumes you only use it in handlers where the `auth` middleware has run. That way when you use it correctly, you don't have to add branching logic for `req.oidc` being undefined.

eg

```js
app.use(auth({...}))

app.get('/my-login', (req, res) => {
  if (!req.oidc) {
    // Why do I have to write this unreachable branch? I know it's defined because of the lines above
  }
  return req.oidc.login();
})
```

You can't use `req.oidc` in a handler that runs before the `auth` middleware, so why would you ever reference it.

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
[...]

PR #398, opened 2022-10-16
fix: change oidc to optional in Request, Response

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:

### Description

I think #397 problem should be solved by changing the `oidc` attribute to optional. Compilation errors can occur in existing code, which can be a kind of breaking change, but it's better than an unpredictable runtime error.

Once this code is applied, everywhere to refer to the `oidc` attribute in TypeScript, we can use it with [Type Guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) like this:

```
if (!req.oidc) {
  throw new Error(`Expects 'oidc` property.`)
}

// After passing the if statement above, we can access the oidc property. ex)
const posts = getUserPosts(req.oidc.user.email)
```

or getUserPosts can handle 'undefined', we can use optional chaning:

```
const posts = getUserPosts(req.oidc?.user.email)
```

### References

- #397 

### Testing

Since it is a simple type change without any code changes, you can test it by checking whether a tsc error occurs when accessing `req.oidc.user` without type guard .

> Describe how this can be tested by reviewers. Be specific about anything not tested and reasons why. If this library has unit and/or integration testing, tests should be added for new functionality and existing tests should complete without errors.
>
> Please include any manual steps for testing end-to-end or functionality not covered by unit/integration tests.
>
> Also include details of the environment this PR was developed in (language/platform/browser version).

### Checklist

    - (There is no TypeScript documentation, so there is nowhere to fix it.)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -183,11 +183,11 @@ interface ResponseContext {
 declare global {
   namespace Express {
     interface Request {
-      oidc: RequestContext;
+      oidc?: RequestContext;
     }
 
     interface Response {
-      oidc: ResponseContext;
+      oidc?: ResponseContext;
     }
   }
 }


adamjmcgrath:
See #397

PR #405, opened 2022-10-27
Update typedoc to latest version

ewanharris:
### Description

This requires updating the typescript version and adding a basic tsconfig that just references the types file. Also corrects readme setting, fixes (most) @link usage that doesn't resolve, removes unused arguments

### Checklist


index.d.ts
```diff
@@ -202,7 +202,7 @@ interface LoginOptions {
   authorizationParams?: AuthorizationParameters;
 
   /**
-   *  URL to return to after login, overrides the Default is {@link Request.originalUrl}
+   *  URL to return to after login, overrides the Default is {@link express!Request.originalUrl Request.originalUrl}
```
ewanharris, line 205: This now links externally to the express docs based on `externalSymbolLinkMappings` in typedoc.js

index.d.ts
```diff
@@ -217,7 +217,7 @@ interface LoginOptions {
  */
 interface LogoutOptions {
   /**
-   *  URL to returnTo after logout, overrides the Default in {@link ConfigParams.routes.postLogoutRedirect routes.postLogoutRedirect}
+   *  URL to returnTo after logout, overrides the Default in {@link ConfigParams.routes routes.postLogoutRedirect}
```
ewanharris, line 220: Parameters in a nested object don't get id's so it's not possible to link to them, so instead changes this to link to `Config.routes` so that it does resolve

adamjmcgrath (approved):
Thanks 👍

PR #414, opened 2022-11-03
Adding client id to _auth_verification cookie key to support multiple…

MatthewBacalakis:
… clients on same protocol and domain.

The key("auth_verification") of the cookie storing nonce/state is problematic in cases where multiple applications share a common protocol and domain.  In that case multiple concurrent logins can conflict with the login for one app clearing the auth_verification cookie intended for the other app.  I recalled the SPA SDK used to have a similar problem and for this reason clientId was added to the key of the nonce/state storage.

To solve this the same way I've appended .CLIENT_ID to the key of the cookie storing nonce.

### DIFF
--- a/test/callback.tests.js
+++ b/test/callback.tests.js
@@ -30,8 +30,8 @@ const defaultConfig = {
 };
 let server;
 
-const generateCookies = (values, clientID) => ({
-  ['auth_verification.' + clientID]: JSON.stringify(values),
+const generateCookies = (values, customTxnCookieName) => ({
+  [customTxnCookieName || 'auth_verification']: JSON.stringify(values),
 });
 
 const setup = async (params) => {
@@ -145,13 +145,10 @@ describe('callback response_mode: form_post', () => {
         body: { err },
       },
     } = await setup({
-      cookies: generateCookies(
-        {
-          nonce: '__test_nonce__',
-          state: '__test_state__',
-        },
-        clientID
-      ),
+      cookies: generateCookies({
+        nonce: '__test_nonce__',
+        state: '__test_state__',
+      }),
       body: true,
     });
     assert.equal(statusCode, 400);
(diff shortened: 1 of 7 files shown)


adamjmcgrath:
Thanks for raising this @MatthewBacalakis!

I know spa-js solves it this way, but I'd like to solve this by allowing the user to override the transaction cookie name themselves.

Users can already set a custom session cookie name with `session.name`, so this would be consistent with the rest of the API. I'd also like to keep the default name the same and offer some flexibility to users that might want to partition these cookies by audience/scope/something else.

We already have `transactionCookie` config, so it would be a case of adding a `name` option to it eg.

```js
app.use(auth({ transactionCookie: { name: 'state.client1' } }));
app.use(auth({ transactionCookie: { name: 'state.client2' } }));
```

MatthewBacalakis:
Updated this so that the cookie can optionally be given any name instead of appending the client id.  If transactionCookie.name is set in config that name is used.  If not the current auth_verification name is used.

test/callback.tests.js
```diff
@@ -287,7 +287,9 @@ describe('callback response_mode: form_post', () => {
         legacySameSiteCookie: false,
       },
       cookies: {
-        _auth_verification: JSON.stringify({ state: '__test_state__' }),
+        ['_auth_verification.']: JSON.stringify({
```
adamjmcgrath, line 290:
```suggestion
        ['_auth_verification']: JSON.stringify({
```

adamjmcgrath (changes requested):
lgtm 👍 

Could you also update the type defs https://github.com/auth0/express-openid-connect/blob/master/index.d.ts#L480

Something like

```ts
transactionCookie?: Pick<CookieConfigParams, 'sameSite'> & { name?: string };
```

adamjmcgrath:
This got released in [2.10.0](https://github.com/auth0/express-openid-connect/releases/tag/v2.10.0) - thanks for your contribution @MatthewBacalakis!

PR #424, opened 2022-11-29
Remove dependency on `cb` lib

### LINKED ISSUE #423: Dependency "cb" causing security issues in Nexus because of missing LICENSE file and being 10 years from last update
### Describe the problem
We are using express-openid-connect for our project to do the whole auth0 process.
But our production build is no longer able to build because Nexus complains about Security vulnaribilities. It is complaing about the dependency "cb" used by express-openid-connect as it is missing a LICENSE file and that the package is older than 5 years, actually the last update was 10 years ago.

Is there any chance to replace this outdated dependency?

My project should be able to use express-openid-connect without running into security vulnerabilities. I expect express-openid-connect to use mantained packages and not totally outdated stuff.

- **Version of this library used:**
LATEST

kmannislands:
### Description

Remove dependency on 10 year old `cb` library. The downside of the supply chain security vector of this lib (especially when used in sensitive middle ware like this) outweighs the trivial utility that `express-oidc-connect` was getting from the dependency. It was being used only to memoize callbacks in 3 locations.

### References
Closes #423 

### Testing

All unit tests passing should cover this change as it is very straight-forward. Code inspection of the `cb` lib indicated that this should behave identically.

### Checklist

- [ ] The correct base branch is being used, if not the default branch

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -1,4 +1,3 @@
-const cb = require('cb');
 const url = require('url');
 const urlJoin = require('url-join');
 const { TokenSet } = require('openid-client');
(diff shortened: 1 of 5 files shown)


adamjmcgrath (approved):
Thanks @kmannislands

PR #435, opened 2023-01-10
Remove unused nonce and code_verifier, do not memoize Issuer.discovery

ewanharris:
### Description

f8e5307e4ba3daed48549dae8ed8b14c665ea523 - Given that `nonce` and `code_verifier` are now generated directly onto `authVerification` there is no need to generate them onto `stateValue` as they are unused

e4f4e5c97e950007c02d1ce9afebdb30c4c3c959 - The return value of get is already memoized so we do not need to memoize this also. Also moves the test setup to use `beforeEach` so that the mock in the top level suite is setup before the client is created (in mocha before runs before beforeEach)

### Checklist

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -205,18 +205,9 @@ class ResponseContext {
       if (typeof stateValue !== 'object') {
         next(new Error('Custom state value must be an object.'));
       }
-      stateValue.nonce = transient.generateNonce();
+
       if (options.silent) {
         stateValue.attemptingSilentLogin = true;
-      }
-
-      const usePKCE =
-        options.authorizationParams.response_type.includes('code');
-      if (usePKCE) {
-        debug(
-          'response_type includes code, the authorization request will use PKCE'
-        );
-        stateValue.code_verifier = transient.generateCodeVerifier();
       }
 
       const validResponseTypes = ['id_token', 'code id_token', 'code'];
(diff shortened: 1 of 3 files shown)


lib/context.js
```diff
@@ -238,13 +229,19 @@ class ResponseContext {
...
       };
 
+      const usePKCE =
+        options.authorizationParams.response_type.includes('code');
       if (usePKCE) {
-        authVerification.code_verifier = transient.generateNonce();
+        debug(
+          'response_type includes code, the authorization request will use PKCE'
+        );
+        authVerification.code_verifier = transient.generateCodeVerifier();
```
ewanharris, line 244: `generateNonce` and `generateCodeVerifier` are [identical](https://github.com/panva/node-openid-client/blob/main/lib/helpers/generators.js#L10-L11) but I changed to the latter here to avoid any potential confusion reading this code

Widcket (approved):

adamjmcgrath (approved):
lgtm

> https://github.com/auth0/express-openid-connect/commit/e4f4e5c97e950007c02d1ce9afebdb30c4c3c959 - The return value of get is already memoized so we do not need to memoize this also.

I'm assuming `get` is memoized because we always use the same pointer to `config`, rather than a new object

```js
const config = { ... };
get(config);
get(config);

// Rather than 
get({ ...config });
get({ ...config });
```

PR #443, opened 2023-02-14
openid-client update from 4.9.1 to 5.4.0

velias:

### Description
`openid-client` dependency update from 4.9.1 to 5.4.0. Subdependency of 4.9.1 has problem, 5.4.0 removes this problem by removing that subdependency. Dependencies are reduced at all in `openid-client` 5.4.0.

### Testing

No change in code itself, only tests has to be updated as new version 
* returns different error messages
* requires accessToken in some calls where not required previously

### Checklist

### DIFF
--- a/test/callback.tests.js
+++ b/test/callback.tests.js
@@ -210,7 +210,7 @@ describe('callback response_mode: form_post', () => {
     assert.equal(statusCode, 400);
     assert.equal(
       err.message,
-      'failed to decode JWT (JWTMalformed: JWTs must have three components)'
+      'failed to decode JWT (Error: JWTs must have three components)'
     );
   });
 
@@ -960,19 +960,12 @@ describe('callback response_mode: form_post', () => {
           scope: 'openid profile email',
         },
         afterCallback: async (req, res, session) => {
-          const userInfo = await req.oidc.fetchUserInfo();
+          const userInfo = {
+            org_id: 'auth_org_123',
+          };
           return { ...session, ...userInfo };
         },
       };
-
-      // userinfo endpoint will be returned to req.oidc.fetchUserInfo
-      const {
-        interceptors: [interceptor],
-      } = nock('https://op.example.com', { allowUnmocked: true })
-        .get('/userinfo')
-        .reply(200, () => ({
-          org_id: 'auth_org_123',
-        }));
 
       const router = auth(authOpts);
       router.get('/session', async (req, res) => {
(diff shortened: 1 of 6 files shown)


test/callback.tests.js
```diff
@@ -960,20 +960,13 @@ describe('callback response_mode: form_post', () => {
           scope: 'openid profile email',
         },
         afterCallback: async (req, res, session) => {
-          const userInfo = await req.oidc.fetchUserInfo();
```
velias, line 963: This call is not possible as it requires `accessToken`, which is not present in the request context yet. As I understand this is not critical part of the test, important is what this calback returns and if it is put into `authSession`. So I added that data directly here.

test/client.tests.js
```diff
@@ -164,7 +164,7 @@ describe('client initialization', function () {
       const handler = sinon.stub().returns([200]);
       nock('https://op.example.com').get('/foo').reply(handler);
       const client = await getClient({ ...config });
-      await client.requestResource('https://op.example.com/foo');
```
velias, line 167: I had to add token here, even its value is not important. But it is required now.

test/login.tests.js
```diff
@@ -563,10 +563,9 @@ describe('auth', () => {
     });
     assert.equal(res.statusCode, 500);
     console.log(res.body.err.message);
-    assert.match(
+    assert.equal(
       res.body.err.message,
-      /^Issuer.discover\(\) failed/,
-      'Should get error json from server error middleware'
```
velias, line 569: Error message returned by the new version is not very nice here, but it is what it is ;-)

test/logout.tests.js
```diff
@@ -95,7 +95,7 @@ describe('logout route', async () => {
     assert.include(
       response.headers,
       {
-        location: `https://op.example.com/session/end?post_logout_redirect_uri=http%3A%2F%2Fexample.org&id_token_hint=${idToken}`,
```
velias, line 98: Logout redirect url contain one more parameter now, but I don't think it's a problem

velias:
Hi @adamjmcgrath , thanks for reply. 

When I ran `yarn audit` in my project two weeks ago, I got info about critical security vulnerability in one of `openid-client` subdependencies, rated on severity 10! I think it was in `got` or some of its subdependencies. I haven't noted that report, and it disappeared next day, so I'm not sure what happened, maybe it went under some embargo, no idea. But I prepared this PR immediately, as having vulnerability rated 10 is really not good ;-)

Anyway, I think it would be really good to update this library, even on price of lost support for older nodejs versions.

adamjmcgrath:
Hi @velias - thanks for sharing that information

I'm not seeing any vulnerable dependencies when I install the latest version of this SDK

image

I absolutely agree that we should update openid-client, and keep all our dependencies up to date. But we only want to do an unplanned major release if we really need to, and currently I see no issues with using openid-client@4 - so closing this.

Thanks again for your contribution. We will definitely upgrade to the latest version of openid-client in our next planned major release.

adamjmcgrath:
Yep - definitely, if you see it pop up again feel free to ping this thread. Thanks for being vigilant! 😄

Issue #426, opened 2022-12-06
Defense-In-Depth Post-Login CSRF Support


madaster97:
### Describe the problem you'd like to have solved
This SDK has one post-login CSRF protection method (`SameSite=Lax` cookies discussed [here](https://github.com/auth0/express-openid-connect/issues/322#issuecomment-1029929634)), but would benefit from additional methods.

I've been using this SDK for prototyping, and only recently realized I had a post-login CSRF attack. For reasons most servers I work with need `Same=None; Secure` in their cookies, and so lose the default CSRF protection from this library.

### Describe the ideal solution
At the very least, I'd like to add a warning when someone sets the `appSession` cookie to `SameSite=None`, reminding them to also add CSRF protection.

Taking that a step further, I'd like this library to offer additional CSRF protections, and potentially enforce them (with an override) when users set `SameSite=None`.

#### CSRF Token in AppSession
I've been pouring over the [OWASP CSRF Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#synchronizer-token-pattern), and I think a reasonable approach that would be easy for this library would be the [Synchronizer Token Pattern](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html), specifically per-session (instead of per-request) tokens. 
> In per-session token implementations after the initial generation of a token, the value is stored in the session and is used for each subsequent request until the session expires. 

Implementing this could look like the following:
- A system for creating + storing per-session CSRF tokens in the `appSession` cookie. We'd rotate this token every time a new session ID is set
- A new request or response method for retrieving the CSRF token from the session. You'd store this in your rendered HTML forms (as a hidden form field) or in the DOM for use in your [AJAX requests](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#javascript-guidance-for-auto-inclusion-of-csrf-tokens-as-an-ajax-request-header).
- A new config option called `csrfCheck`, which would be a function expected to return the CSRF token supplied in the request. The function would have the request headers and body (assumed x-www-form-urlencoded) from the incoming request as inputs, since limiting to these mitigates the risk of insecure requests that can leak the token (via query params).

The `csrfCheck` could be used in a few different ways, such as:
[...]

Issue #332, opened 2022-02-16
Interoperability with express-session


nponeccop:
### Describe the problem you'd like to have solved

I have a legacy app which passes Express session cookies over non-HTTP connections (think websockets or backend message queues). The app used `everyauth` which has been unmaintained for many years now. So I needed to use another social login library/service. But `express-openid-connect` didn't work for me because while it could use Express session stores, it doesn't use Express session cryptography or Express session cookie format, but instead implements its own `apiSession.js` session library, with JOSE/HKDF crypto instead of simple HMAC in `express-session`, symbols to avoid `req` pollution and so on.

### Describe the ideal solution

- implement an `express-session` compatibility mode where the existing Express infrastructure around sessions is used
- Extract `appSession.js` into a separate library/middleware which is used on its own as an alternative to `express-session` if someone wants a better cryptography or more features

## Alternatives and current work-arounds

- hack/fork `express-openid-connect` to use `express-session` and/or avoid extensive monkey-patching
- use `@okta/oidc-middleware` with Auth0

adamjmcgrath:
Hi @nponeccop - thanks for raising this

It would require a considerable refactor to make this SDK compatible with `express-session` and we believe that the built in app session is sufficient.

Could you tell me more about what specifically is stopping you from using the built in session.

>  with JOSE/HKDF crypto instead of simple HMAC in express-session

Why is this an issue for you, and how does it prevent you from using the SDK?

> symbols to avoid req pollution and so on.

Which properties are we adding to `req` that `express-session` is avoiding using Symbols?

adamjmcgrath:
Closing due to innactivity

PR #453, opened 2023-03-28
Add CSRF Token Example

madaster97:
### Description
Building off the discussion in [issue 426](https://github.com/auth0/express-openid-connect/issues/426), I'm adding some documentation (via an example) of adding a Cross-Site Request Forgery (CSRF) Token to this SDK's session and using it to protect a stateful route from CSRF. I also did some quick clean up to the EXAMPLES doc to fix hyperlinks.

In the docs for the example, I call out the fact that you shouldn't really need this, since the SDK's `SameSite=Lax` cookie default provides good protection. However, there are still use cases (for example, sites in iframes requiring `SameSite=None`) that lose that protection, and I'd like to have other mitigations documented.

In the example app, I implemented the [Synchronizer Token](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#synchronizer-token-pattern) pattern, which relies on storing a token in a server session that the server can return to the client and validate later.

#### Token Storage Options
For storing the token, I landed on reading/writing from `req.appSession` directly, after trying a couple other options:
- Trying to add my own, separate session cookie with express-session. I ran into issues, and it sounds like this [isn't supported](https://github.com/auth0/express-openid-connect/issues/332#issuecomment-1042893092)
- Trying to read the `appSession` cookie value directly, and using a hash of the value as the CSRF token. I looked around, and couldn't find a way to retrieve this value from a request.

#### Considered But Not Done
- Writing an example for a single-page-app (SPA), that uses a [cookie-to-header](https://en.wikipedia.org/wiki/Cross-site_request_forgery#Cookie-to-header_token) token like Angular does. I used a conventional server-rendered HTML app with a form POST navigation (rather than fetch).
- Having request-specific CSRF tokens. I used session-specific ones that rotate appropriately based on logout and login events.

The example I wrote has most of the plumbing in place, and it would be easy enough to edit for either use case above.

### Testing
As part of the docs I wrote a sample HTML page that tries to exploit the example's CSRF-protected endpoint, and it appropriately fails!

### Checklist
- [X] All active GitHub checks for tests, formatting, and security are passing
- [X] The correct base branch is being used, if not the default branch


madaster97:
Sad, I thought you might say that! Yep I just took out the edits other than the doc clean up.

In the type documentation, I had this blurb added to the SameSite property docs:
_When setting to 'None' (uncommon), you should implement CSRF protection on your own routes_

What do you think about keeping this in?

adamjmcgrath (approved):
lgtm - and thanks for fixing the numbering on the examples @madaster97!

PR #456, opened 2023-03-29
Typing fix

evertdespiegeleer:
As shown in [one of your examples](https://github.com/auth0/express-openid-connect/blob/master/examples/custom-routes.js#L16), the routes.callback property in the middleware should be able to be equal to `false`. However, the type definitions currently don't allow this.


adamjmcgrath (approved):
Thanks @evertdespiegeleer!

PR #483, opened 2023-09-01
Add loginRedirectCallback

gyx1000:
### Description

> Some frameworks, such as Remix, after generating the first rendering of the website on the backend, use a custom navigation for the page change on the client.
> In this case, an http 302 is not possible, as it will be misinterpreted and does not correspond to the framework flow. To deal with this situation, an optional parameter in the login can be used to customize the redirect response on the authorization url.
>
> No breaking change is involved in this change, as the parameter is optional and the default flow is retained.

### References

> - [Remix redirect response handling](https://github.com/remix-run/remix/blob/main/packages/remix-server-runtime/server.ts#L140)

### Testing

> I've created an example project for testing purposes. Here's how to test it:
>

```bash
git clone https://github.com/gyx1000/remix-express-oidc-redirect
cd remix-express-oidc-redirect
npm i
#
# Configure .env with :
COOKIE_SECRET=
CLIENT_ID=
ISSUER=

npm run dev
```
>
> You can then go to http://localhost:3000 and click on the dashboard link. An error will occur, as Remix will not be able to interpret the 302 response.
> If you install the modified express-openid-connect version of the PR, the authentication process will work.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -226,6 +226,25 @@ interface LoginOptions {
    *  Used by {@link ConfigParams.attemptSilentLogin} to swallow callback errors on silent login.
    */
   silent?: boolean;
+
+  /**
+   * Function for custom redirect response handling.
+   * This can be used for handling login redirection response different from default 302 response
+   *
+   * ```js
+   * app.get('/login', (req, res) => {
+   *  res.oidc.login({ loginRedirectCallback: async (res, authorizationUrl) => {
+   *    res.set('x-custom-header', 'custom-header-value');
+   *    res.set('x-custom-authorization-url', authorizationUrl);
+   *    res.status(204).send();
+   *  });
+   * });
+   * ```
+   */
+  loginRedirectCallback?: (
+    res: Response,
+    authorizationUrl: string
+  ) => Promise<void>;
 }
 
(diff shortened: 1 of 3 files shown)


adamjmcgrath:
Thanks for raising this @gyx1000 - I'll try and take a look at it this week or next week

adamjmcgrath:
Hi @gyx1000 - I've done a bit of research on this and it looks like you have a couple of options to resolve this issue coming in the next version or Remix (see https://github.com/remix-run/react-router/discussions/9859#discussioncomment-6620496) and a workaround to use for the time being (see https://github.com/remix-run/remix/issues/3765#issuecomment-1400921625). 

I'm reluctant to add this hook just for that use case, so closing

gyx1000:
Hi @adamjmcgrath ,

Thank you very much for your time and feedback.
Indeed, I can wait for the remix feature to allow me to manage this situation.

I had done this PR anyway because I don't want to manage authentication in remix but leave it to the server.js express adapter. The adapter serves other routes that aren't remix, and the use of express-openid-connect suited me perfectly.
I fully understand your reluctance to use this hook for this use case.

Thank you again.

PR #670, opened 2025-07-17
fix: #624 fix calculateExp functionality

aks96:
### Description
This PR fixes the session expiry calculation in appSession.js when rolling: true and absoluteDuration: false are set in the session configuration. Previously, sessions would expire immediately due to incorrect expiry logic in the calculateExp function. Now, the session will correctly expire only after the defined rollingDuration period of inactivity.

### Changes
- Updated calculateExp to use only rollingDuration when absoluteDuration is falsy.
- No changes to tests were required; existing tests now pass with the correct logic.

### Impact

- Users with rolling: true and absoluteDuration: false will have sessions that persist for the expected rolling duration.
- No breaking changes or version bumps.

### Related Issues
Fixes session expiry bug described in #624

### DIFF
--- a/lib/appSession.js
+++ b/lib/appSession.js
@@ -91,9 +91,11 @@ module.exports = (config) => {
       return iat + absoluteDuration;
     }
 
-    return Math.min(
-      ...[uat + rollingDuration, iat + absoluteDuration].filter(Boolean)
-    );
+    if (!absoluteDuration) {
+      return uat + rollingDuration;
+    }
+
+    return Math.min(uat + rollingDuration, iat + absoluteDuration);
   }
 
   function setCookie(


lib/appSession.js
```diff
@@ -91,9 +91,11 @@ module.exports = (config) => {
...
 
-    return Math.min(
-      ...[uat + rollingDuration, iat + absoluteDuration].filter(Boolean)
-    );
+    if (!absoluteDuration) {
+      return uat + rollingDuration;
+    }
+
+    return Math.min(uat + rollingDuration, iat + absoluteDuration);
   }
```
kushalshit27, line 100:
```suggestion
function calculateExp(iat, uat) {
  if (!rollingEnabled) {
    return iat + absoluteDuration;
  }

  const rollingExp = uat + rollingDuration;
  const absoluteExp = iat + absoluteDuration;
  
  const validExpirations = [rollingExp, absoluteExp].filter(Boolean);
  return Math.min(...validExpirations);
}
```

aks96: Here filter (Boolean) won't remove iat if its greater than zero and we will encounter with same bug and when absoluteDuration is false then iat will be only number to get with. What your thoughts on this @kushalshit27

kushalshit27: Agree

PR #677, opened 2025-07-22
fix: async functions that require callbacks in promisify utility

kushalshit27:
### Description

This pull request enhances the `safePromisify` function in `lib/utils/promisifyCompat.js` to better handle various types of asynchronous methods. The most important change is the addition of logic to support methods that are both asynchronous and callback-based.

### Testing

- [x] This change passes the existing test

### Checklist

- [ ] The correct base branch is being used, if not the default branch


kishore7snehil (approved):
LGTM!

PR #704, opened 2025-09-04
Add support for MRRT

martinml:
### Description

This PR adds support for [Auth0's MRRT](https://auth0.com/docs/secure/tokens/refresh-tokens/multi-resource-refresh-token) feature.

High-level overview of the changes this PR includes:

1. Audience and organization are now saved within the tokenset.
1. A "token history" feature that keeps inside the user session a list of previous but non-expired tokens. This is needed so the SDK can find "compatible" tokens when doing a MRRT refresh (i.e. if I have a token for audience `A` and I need a token for audience `B`, I can do a refresh from `A` to `B`). Can be enabled with the new `tokenHistory` option, but it's disabled by default.
1. A refactor of the existing middlewares signatures so they accept an object instead of discrete arguments. This was made so routes can specify an `authorizationParams` argument that will override the one set at the SDK level. This is not a breaking change since I included a compatibility layer that detects old vs. new arguments and normalizes them (see `requiresAuthLegacyArgs.js`).
1. A `useMrrt` option, to enable the potential usage of MRRT, depending on whatever it's in the token history. It requires `tokenHistory` to be enabled and [proper configuration in Auth0](https://auth0.com/docs/secure/tokens/refresh-tokens/multi-resource-refresh-token/configure-and-implement-multi-resource-refresh-token). Disabled by default.
1. An `autoRefreshExpired` option, to improve the DX when using MRRT. This way, requesting a series of tokens with different audiences/scopes to protect a route is a declarative operation, instead of having to check whether tokens are expired.
1. A new cookie format. The previous one was storing the session data directly into the root of the data structure, which made impossible to store "sibling" properties. Old cookies are transparently converted into the new version, and a versioning schema is introduced. Note that this only applies when _not_ using a custom store: we _can_ store other properties adjacent to the session data one.

### Testing

### Checklist

### DIFF
--- a/lib/tokenSetUtils.js
+++ b/lib/tokenSetUtils.js
@@ -20,19 +20,17 @@ const TokenSetUtils = {
       refresh_token: tokenSet.refresh_token,
       token_type: tokenSet.token_type,
       expires_at: tokenSet.expires_at,
+      audience: tokenSet.audience,
+      scope: tokenSet.scope,
+      organization: tokenSet.organization,
     };
 
-    if (tokenSet.audience) {
-      output.audience = tokenSet.audience;
-    }
-
-    if (tokenSet.scope) {
-      output.scope = tokenSet.scope;
-    }
-
-    if (tokenSet.organization) {
-      output.organization = tokenSet.organization;
-    }
+    // Don't keep optional values that are nullish.
+    Object.keys(output).forEach((key) => {
+      if (!output[key]) {
+        delete output[key];
+      }
+    });
 
     return output;
   },


end-to-end/fixture/helpers.js
```diff
@@ -8,6 +8,9 @@
 
 const baseUrl = 'http://localhost:3000';
 
+// Avoid breached password warnings from Chrome that make tests flaky.
+const password = `rnd-pwd-${Math.random()}`;
```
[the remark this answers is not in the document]
martinml: Not a real security context. Just dummy password for tests.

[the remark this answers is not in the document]
martinml: However, I will change this to avoid false positives 👍

lib/tokenSets.js
```diff
@@ -0,0 +1,167 @@
...
+   * with the provided `authorizationParams` and is active (i.e. not expired).
+   *
+   * @param {import('..').TokenSetParameters[]} tokenSets
+   * @param {import('..').AuthorizationParameters} authorizationParams
+   * @returns {import('..').TokenSetParameters | undefined}
+   */
+  findCompatibleActive(tokenSets, authorizationParams) {
+    return tokenSets.find(
+      (ts) =>
+        !TokenSetUtils.isExpired(ts) &&
```
aks96, line 21: these 3 methods findCompatibleActive, findCompatibleExpired findCompatibleRefreshable looks similar. Could we refactor this?

lib/tokenSets.js
```diff
@@ -0,0 +1,167 @@
...
+    const tokenSets = this.getAvailableTokenSets(req);
+
+    const mergedParams = {
+      ...config.authorizationParams,
+      ...authorizationParams,
+    };
+
+    /** @type {import('..').TokenSetParameters | undefined} */
+    let found;
+
```
aks96, line 114: Shouldn't we consider the most recent token? Could you re-check this once.

end-to-end/attempt-silent-login.test.js
```diff
@@ -96,8 +97,13 @@ describe('attempt silent login', async () => {
...
 
-    const loggedOutCookies = await context.cookies(baseUrl);
-    assert.isNotOk(loggedOutCookies.find(({ name }) => name === 'appSession'));
+    /**
+     * Logged out cookies will still have something inside them, and we can't
+     * verify their content since they're encrypted, so at least let's check it's
+     * small enough that a token can't possibly fit inside.
+     */
+    const loggedOutCookies = await context.cookies();
+    assert.isTrue(loggedOutCookies.find(({ name }) => name === 'appSession').size < 200);
```
aks96: here if loggedOutCookies.find does find the cookie it returns `undefined`
Shall we do something 
`const appSessionCookie = loggedOutCookies.find(({ name }) => name === 'appSession');
assert.isTrue(!appSessionCookie || appSessionCookie.size < 200);`

martinml: I think this is fixed now [here](https://github.com/auth0/express-openid-connect/pull/704/commits/95d45591c85c4df3b6647702d9ce4854c6ae06fa). Some code was not expecting the possibility of optional props to actually be present in the tokenset but with a nullish value. This caused the cookie not to be properly cleared.

PR #760, opened 2026-01-09
feat: Test infrastructure for v6

aks96:
### Description 
This repo has open PR for node.js 20+ and ESM only which all existing test passing. The next phase for OIDC migration v4 to v6. Before changing runtime behaviour this PR has setup test infrastructure for next steps.

### Changes 

- Refactors and centralizes OIDC-related test helpers and mocks
- Reduces coupling between tests and OIDC v4–specific implementation details
- Removes assumptions tied to legacy CommonJS behavior
- Simplifies token, issuer, and client stubbing to be compatible with upcoming OIDC v6 changes
**Does not change any behaviour of repo**


cschetan77:
#785

PR #785, opened 2026-02-09
migration: openid-client and jose migration

aks96:
### Summary
Upgrades `openid-client` from v4 to v6 and `jose` from v2 to v6. These are now ESM-only packages, which drives the Node.js version requirement change.
The core middleware API, session handling, and all authentication methods remain unchanged.

### Breaking Changes
See [V3_MIGRATION_GUIDE.md](./V3_MIGRATION_GUIDE.md) for before/after examples and migration steps for each change.

  1. **Node.js version** — requires `^20.19.0 || ^22.12.0 || >= 23.0.0`
  2. **`httpAgent` config removed** — replaced by `customFetch`. Passing `httpAgent` throws at startup.
  3. **`clientAssertionSigningAlg` now required** — the implicit `RS256` default has been removed. Required when `clientAssertionSigningKey` is a PEM, Buffer, KeyObject, or JWK without an `alg` property.
  4. **`ES256K` and `EdDSA` removed** from valid `clientAssertionSigningAlg` values. Use `Ed25519` instead of `EdDSA`. `ES256K` has no replacement.
  5. **`afterCallback` behavior change** — `req.oidc` now reflects the incoming user's new tokens during the callback, not the previous session state.
  6. **Session cookie dropped on streaming responses** — in v2, the session cookie was injected just before headers were flushed, so it worked even on streaming responses. In v3, the cookie is written at `res.end()` time instead. If your route calls `res.write()`, `res.flushHeaders()`, or `res.writeHead()` before the response ends (e.g. SSE, chunked responses), the session cookie will be silently skipped. If you rely on token rotation, avoid using it on streaming routes.
  7. **`clientAssertionSigningKey` type updated** — if you use TypeScript, the accepted type for `clientAssertionSigningKey` has changed to align with jose v6 and the Web Crypto API. In practice, all previously supported key formats (PEM string, Buffer, KeyObject, JWK) still work at runtime — only the TypeScript type definition has changed

#### Manual testing against a real Auth0 tenant

  | Scenario | Flow |
  |---|---|
  | Basic login flow | Discovery, token exchange, callback URL handling, session cookie written correctly |
  | `customFetch` wiring | Called for both discovery and token endpoints, confirmed via logging |
  | `httpAgent` rejection | Throws `TypeError: "httpAgent" is not allowed` at startup before the app boots |
  | Private key JWT with PEM + `clientAssertionSigningAlg: 'RS256'` | Full login succeeds using signed `client_assertion` JWT |
  | Private key JWT without `clientAssertionSigningAlg` | Throws at startup |
  | `afterCallback` new behavior | `req.oidc.user` reflects the incoming user's tokens, confirmed `sub` matches the new ID token |
  | Streaming response (`res.writeHead` before `res.end`) | No crash, session cookie silently skipped. |

#### Unit tests
Unit tests have been updated to reflect the new API and behaviour.

#### End-to-end tests
The existing end-to-end test suite covers the full login/logout flow, token refresh, backchannel logout (basic, re-login after logout, custom genid, custom query store), and custom session store scenarios using a local OIDC provider.


V3_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,190 @@
+# V3 Migration Guide
```
cschetan77, line 1: Just a thought. 
Since consumer facing api has no breaking changes, do we need to mention that and it's usage examples here, in detail.  
IMO, if consumer comes to read the migration guide, consumer has to go through a lot of detail to be assured that there are no breaking changes. 
Would it not suffice to just mention the breaking changes and what consumer are required to follow.
@aks96 What do you think!

aks96: Here we are making sure user understand what changes has been done, regarding breaking changes it just user has to update node, openid-client and jose version.

lib/context.js
```diff
@@ -368,55 +380,160 @@ class ResponseContext {
...
-          },
-          clientAssertionPayload: {
-            aud: issuer.issuer,
-          },
-        });
+        const responseType = config.authorizationParams.response_type;
+
+        // Build a Request object for openid-client v6
+        const protocol = req.protocol;
+        const host = req.get('host'); // Use req.get('host') to include port
```
panva: This will not work in any deployment behind a TLS terminating reverse proxy. It'll give back the internal hostname and port which has no relation to the URL the browser is hitting. This URL is used by openid-client to populate the `redirect_uri` access token request parameter, thus causing a mismatch.

panva: Using req getters is necessary to account for the express application's `trust proxy` settings, https://expressjs.com/en/guide/behind-proxies.html

panva: Unfortunately there's a bug in express@4 where req.host is an alias to req.hostname and thus not containing ports. express@5 req.host works as expected.

To work around this use the following (taken from https://github.com/panva/openid-client/blob/b77d87c1e2f5fef6fab501de615fb83a74a0251f/src/passport.ts#L211C1-L232)

```ts
function host(req: express.Request): string | undefined {
  try {
    const trust = req.app.get('trust proxy fn')
    let val = req.get('x-forwarded-host')

    if (!val || !trust(req.socket.remoteAddress, 0)) {
      val = req.get('host')
    } else if (val.indexOf(',') !== -1) {
      val = val.substring(0, val.indexOf(',')).trimRight()
    }

    return val || undefined
  } catch {
    return req.host
  }
}
```

lib/context.js
```diff
@@ -368,55 +380,160 @@ class ResponseContext {
...
+        const currentUrl = new URL(`${protocol}://${host}${req.originalUrl}`);
```
panva: To account for mounted app scenarios

```suggestion
        const currentUrl = new URL(`${protocol}://${host}${req.originalUrl ?? req.url}`);
```

lib/client.js
```diff
@@ -17,35 +17,168 @@ function sortSpaceDelimitedString(string) {
...
+    clientAuth,
+    discoveryOptions,
+  );
+
+  // Get server metadata for validation
+  const serverMetadata = configuration.serverMetadata();
 
   const issuerTokenAlgs = Array.isArray(
-    issuer.id_token_signing_alg_values_supported,
+    serverMetadata.id_token_signing_alg_values_supported,
```
cschetan77, line 156: `id_token_signed_response_alg` is missing from `clientMetadata`

```javascript
    const clientMetadata = {
    [client.clockTolerance]: config.clockTolerance,
    id_token_signed_response_alg: config.idTokenSigningAlg,
  };
```

lib/client.js
```diff
@@ -17,35 +17,168 @@ function sortSpaceDelimitedString(string) {
...
+
+  if (responseType === 'id_token') {
+    execute.push(client.useIdTokenResponseType);
+  } else if (responseType === 'code id_token') {
+    execute.push(client.useCodeIdTokenResponseType);
+  }
+  // 'code' is the default in v6, no special execute function needed
+
+  return execute;
+}
```
cschetan77: I see 2 problems with this design - 
1. The function getExecuteFunctions is named and designed to encapsulate the logic of which execute functions are needed, but the caller then mutates the returned array to add support for `allowInsecureRequests` later. 
2. function name could be something better, maybe - `getDiscoveryExecuteFunctions` or `buildDiscoveryExecute` 


```suggestion
function getExecuteFunctions(config) {
  const execute = [];
  const responseType = config.authorizationParams.response_type;

  if (responseType === 'id_token') {
    execute.push(client.useIdTokenResponseType);
  } else if (responseType === 'code id_token') {
    execute.push(client.useCodeIdTokenResponseType);
  }
  if (new URL(config.issuerBaseURL).protocol === 'http:') {
      execute.push(client.allowInsecureRequests);
  }
  // 'code' is the default in v6, no special execute function needed
  return execute;
}
```

lib/client.js
```diff
@@ -86,66 +219,71 @@ async function get(config) {
...
-    },
-    jwks,
-  );
-  applyHttpOptionsCustom(client);
-  client[custom.clock_tolerance] = config.clockTolerance;
-
+  // Handle Auth0-specific logout
+  let auth0Logout = false;
   if (config.idpLogout) {
+    const issuerUrl = url.parse(serverMetadata.issuer);
```
cschetan77: `url.parse()` is deprecated [DEP0169](https://github.com/nodejs/node/issues/61724)
Better we update these calls to WHATWG `URL` API as well, since we are already using it to construct `issuerUrl` for discovery.
Same in `buildEndSessionUrl` function.

lib/client.js
```diff
@@ -86,66 +219,71 @@ async function get(config) {
...
+      ...extraParams,
+      returnTo: post_logout_redirect_uri,
+      client_id: config.clientID,
+    };
+
+    Object.entries(parsedUrl.query).forEach(([key, value]) => {
+      if (value === null || value === undefined) {
+        delete parsedUrl.query[key];
+      }
+    });
```
cschetan77: I think this filtering is unnecessary. `filteredParams` already contains sanitized entries and `parsedUrl.query` is constructed out of it, every value in parsedUrl.query is guaranteed non-null.

lib/client.js
```diff
@@ -17,35 +17,168 @@ function sortSpaceDelimitedString(string) {
...
+    typeof keyData === 'object' && !Buffer.isBuffer(keyData)
+      ? { key: keyData, format: 'jwk' }
+      : keyData,
+  );
+
+  // Export as PKCS8 DER and import to Web Crypto API CryptoKey
+  const pkcs8Der = keyObject.export({ type: 'pkcs8', format: 'der' });
+
+  return crypto.subtle.importKey('pkcs8', pkcs8Der, algorithm, false, ['sign']);
+}
```
cschetan77: `importPKCS8AsCryptoKey` reimplements key import logic that `jose@6` (already a dependency) provides out of the box. 
The current approach also does a manual roundtrip — `input -> Node.js KeyObject -> PKCS8 DER bytes -> Web Crypto CryptoKey` — and maintains a hard-coded algorithm map to do it, this can all be avoided by using `jose`. 

Two concrete problems with this:

  1. **EdDSA is silently unsupported.** - The algorithm map covers 9 algorithms but not `EdDSA` (Ed25519/Ed448), which both `jose` and `openid-client` v6 support. A user configuring `clientAssertionSigningAlg: 'EdDSA'` gets a hard `Unsupported algorithm` error.

  2. **Low-level crypto plumbing in a library is a huge unnecessary risk.** - Low-level crypto code is the most sensitive code in a library. Every line of it is a potential vulnerability surface. Most of the manual `crypto.subtle` code can be replaced by a well-tested library.


`jose` v6 exports `importPKCS8` and `importJWK` which return a `CryptoKey` directly — exactly what `PrivateKeyJwt()` needs:

  ```js
  const { importPKCS8, importJWK } = require('jose');

  const privateKey = (typeof keyData === 'object' && !Buffer.isBuffer(keyData))
    ? await importJWK(keyData, alg)
    : await importPKCS8(keyData.toString(), alg);

  return client.PrivateKeyJwt(privateKey, { algorithm: alg });
  ```

This drops importPKCS8AsCryptoKey entirely and keeps crypto handling inside a dependency that already owns that responsibility.

cschetan77 (changes requested):

lib/crypto.js
```diff
@@ -49,25 +49,25 @@ if (crypto.hkdfSync) {
...
-      current = key;
-    }
-    keystore.add(key);
-  });
+  const keys = secrets.map((secretString) =>
+    forEncryption ? encryption(secretString) : signing(secretString),
+  );
+  const current = keys[0];
+  // Return keys as an array (or single key if only one)
+  const keystore = keys.length === 1 ? current : keys;
```
cschetan77: This ternary in `getKeyStore` serves no purpose other than creating a type inconsistency, jose v6 doesn't need a **KeyStore** concept anymore. Each caller now handles multi-key iteration itself via a loop, so the distinction between a single key and an array has no meaning at the getKeyStore level.
This inconsistency has to be  then handled in every caller with-
```javascript
const keys = Array.isArray(keystore) ? keystore : [keystore];
```
Suggest always returning an array as before
```javascript
return [current, keys];
```
This removes the ternary here and eliminates the `Array.isArray` normalisation in all callers.

lib/appSession.js
```diff
@@ -1,11 +1,11 @@
 const { strict: assert, AssertionError } = require('assert');
 const {
-  JWE,
+  CompactEncrypt,
+  compactDecrypt,
   errors: { JOSEError },
 } = require('jose');
 const safePromisify = require('./utils/promisifyCompat');
 const cookie = require('cookie');
-const onHeaders = require('on-headers');
```
cschetan77, line 8: `on-headers` dep can be removed from package json

lib/context.js
aks96: addressed this issue

lib/appSession.js
```diff
@@ -171,19 +186,19 @@ module.exports = (config) => {
...
-      const { protected: header, cleartext } = decrypt(idOrVal);
+      const { protected: header, cleartext } = await decrypt(idOrVal);
       return {
         header,
         data: JSON.parse(cleartext),
       };
     }
 
-    getCookie(req) {
+    async getCookie(req) {
```
cschetan77, line 196: This doesn't require to be `async`.

lib/appSession.js
```diff
@@ -387,16 +407,16 @@ module.exports = (config) => {
...
-        store.setCookie(req[REGENERATED_SESSION_ID] || id, req, res, { iat }),
-      );
-
-      const { end: origEnd } = res;
       res.end = async function resEnd(...args) {
         try {
+          await store.setCookie(req[REGENERATED_SESSION_ID] || id, req, res, {
+            iat,
+          });
           await store.set(id, req, res, {
```
cschetan77: Little doubtful about the order of these 2 calls here. 
What if `store.set` fails to persist data and throws an error, but the response does contains the Set-Cookie header. It sends the session cookie pointing to data that was never persisted. The client ends up with a broken, unresolvable session cookie. 
We should swap the order, store.set first, then store.setCookie, then origEnd. 
If `store.set` fails, the original code sends no cookie.

lib/appSession.js
```diff
@@ -98,7 +113,7 @@ module.exports = (config) => {
     return Math.min(uat + rollingDuration, iat + absoluteDuration);
   }
 
-  function setCookie(
+  async function setCookie(
```
cschetan77, line 116: Should we also set `res.headersSent` guard in this function as well, in similar lines to what was added to `CustomStore.setCookie` !

cschetan77 (changes requested):

lib/context.js
```diff
@@ -220,8 +223,17 @@ class RequestContext {
...
+    const ts = tokenSet.call(this);
+
+    if (!ts || !ts.access_token) {
+      throw new Error(
+        `Access token is required to fetch user info but none was found in the session.\n`,
+      );
+    }
+
+    const claims = ts.claims();
+    return oidcClient.fetchUserInfo(configuration, ts.access_token, claims.sub);
```
cschetan77: if `TokenSet.claims()` returns `undefined` when id_token is absent, the `claims.sub` would throw a TypeError and crash out. 
We should put a check

```suggestion
    const claims = ts.claims();
    return oidcClient.fetchUserInfo(configuration, ts.access_token, claims?.sub);
```

lib/context.js
```diff
@@ -519,39 +531,161 @@ class ResponseContext {
...
+      const wasAuthenticated = req.oidc.isAuthenticated();
+      const previousSub = wasAuthenticated ? req.oidc.user.sub : null;
 
       if (config.afterCallback) {
+        // Temporarily set the session so that req.oidc methods can access token data
+        const originalSession = req[config.session.name]
+          ? { ...req[config.session.name] }
+          : {};
+        Object.assign(req[config.session.name], session);
+
```
cschetan77: Isn't this a kind of breaking change. 
`req.oidc` inside afterCallback reflected the old session.Any existing `afterCallback` implementation that relied on this, checking the previous user's state, reading old tokens would now see completely different behavior.
This is a behavioral change to req.oidc inside a public hook and should be called out. 

Also, if after the temporary mutation, the `await config.afterCallback()` throws an error, the execution will not restore the original state. We should handle it via try/finally.

lib/context.js
```diff
@@ -469,13 +480,14 @@ class ResponseContext {
     let { config, req, res, transient, next } = weakRef(this);
     next = once(next);
     try {
-      const { client, issuer } = await getClient(config);
+      const { configuration } = await getClient(config);
+      // eslint-disable-next-line no-unused-vars
       const redirectUri = options.redirectUri || this.getRedirectUri();
```
cschetan77: Shouldn't we pass `redirectUri` explicitly in endpointParameters.
Also, `options.redirectUri` override will be silently ignored.

cschetan77 (changes requested):

lib/context.js
```diff
@@ -519,39 +531,161 @@ class ResponseContext {
...
       }
 
-      let session = Object.assign({}, tokenSet); // Remove non-enumerable methods from the TokenSet
-      const claims = tokenSet.claims();
+      // Build session from token response
+      let session = {
+        id_token: tokenResponse.id_token,
+        access_token: tokenResponse.access_token,
+        refresh_token: tokenResponse.refresh_token,
+        token_type: tokenResponse.token_type,
```
cschetan77: The custom `TokenSet` class normalizes the token_type to 'Bearer'. 
But both `callback()` and `refresh()` bypass TokenSet and write directly to the session.
`req.oidc.accessToken.token_type` is unaffected since it always goes through `TokenSet` at read time. But the session itself stores the raw value, which seems to create inconsistencies in 2 cases-
1. afterCallback receives session.token_type === 'bearer'  which was 'Bearer' in v4.
2. Persisted sessions store lowercase, any server-side code reading the session directly sees a different value than v4
We should normalize at write time in both places 
```javascript
const normalizeTokenType = (t) => t?.toLowerCase() === 'bearer' ? 'Bearer' : t;

  // callback()
  token_type: normalizeTokenType(tokenResponse.token_type),

  // refresh()
  token_type: normalizeTokenType(newTokenSet.token_type),
```

lib/context.js
```diff
@@ -601,14 +735,58 @@ class ResponseContext {
...
+      if (jwksCache && jwksCache.jwks) {
+        JWKS = createLocalJWKSet(jwksCache.jwks);
+      } else if (serverMetadata.jwks_uri) {
+        // Fallback: fetch JWKS if not cached
+        const jwksResponse = await fetch(serverMetadata.jwks_uri);
+        if (!jwksResponse.ok) {
+          throw new Error('Failed to fetch JWKS');
+        }
+        const jwks = await jwksResponse.json();
+        JWKS = createLocalJWKSet(jwks);
```
cschetan77: I see 3 problems here, not very sure if the last one is a real concern - 
1. No caching in the fallback path - `createLocalJWKSet` produces a static snapshot. Every backchannel logout request that misses `getJwksCache` makes a fresh uncached HTTP request.
2. Raw fetch fallback bypasses the SDK's custom fetch - 
`const jwksResponse = await fetch(serverMetadata.jwks_uri);  // global fetch`
This bypasses `createCustomFetch` configured in `client.js`,  no User-Agent, no Auth0-Client telemetry, no httpTimeout. A hanging JWKS request will never time out.
3. `getJwksCache` accesses internal library state -
  `const jwksCache = oidcClient.getJwksCache(configuration);
  // internally: int(config).jwksCache`,  private symbol-keyed store
 While exported, it doesn't seem to be a stable public API. It also returns `undefined` on every fresh process start or after key rotation (checked via cache.uat), meaning the fallback seems to hit constantly in practice (please correct if this is not the case).

Can't we use `createRemoteJWKSet` from `jose`, configuring it with customFetch that we have - 
```javascript
const JWKS = createRemoteJWKSet(new URL(serverMetadata.jwks_uri), {
    [customFetch]: customFetch,
  });
```
Let me know your thoughts.

cschetan77 (changes requested):

lib/context.js
```diff
@@ -469,13 +495,14 @@ class ResponseContext {
     let { config, req, res, transient, next } = weakRef(this);
     next = once(next);
     try {
-      const { client, issuer } = await getClient(config);
-      const redirectUri = options.redirectUri || this.getRedirectUri();
+      const { configuration } = await getClient(config);
+      // Note: In openid-client v6, redirect_uri is automatically extracted from the
```
nandan-bhat: `redirect_uri` is still part of the public API and the typedocs still show `res.oidc.callback({ redirectUri: ... })`. 

If we now always derive the redirect URI from the current request URL, apps with custom callback routes or multiple domains may send a different `redirect_uri` to the token endpoint than the one used during login.

If this is intentional, can we call it out as a breaking change and update the `types/docs` ?

lib/context.js
```diff
@@ -519,39 +546,166 @@ class ResponseContext {
...
+        }
+
+        // Don't include standard ports in URL
+        const standardPorts = { http: '80', https: '443' };
+        const needsPort = port && port !== standardPorts[protocol];
+
+        // Build host with port if needed
+        const host = needsPort ? `${hostname}:${port}` : hostname;
+
+        const currentUrl = new URL(`${protocol}://${host}${req.originalUrl}`);
```
nandan-bhat: IMO, we should not drop `options.redirectUri` support here. Even though `openid-client` v6 can infer the `redirect_uri` from the URL passed to `authorizationCodeGrant()`, this SDK can still decide which URL to pass in.

To preserve the existing `express-openid-connect` API, we should use `options.redirectUri` when it is supplied, and only infer the callback URL from the request when it is not. 

Otherwise, existing apps that rely on `res.oidc.callback({ redirectUri })`  may break silently, even though the option is still documented and typed.

lib/appSession.js
```diff
@@ -387,19 +412,21 @@ module.exports = (config) => {
       attachSessionObject(req, sessionName, {});
     }
 
+    const { end: origEnd } = res;
+
     if (isCustomStore) {
       const id = existingSessionValue || (await generateId(req));
```
nandan-bhat, line 423: Can we add coverage for responses that flush headers before `res.end()` ?. Since session cookies are now written from the `res.end wrapper`, `setCookie()` may skip writing the cookie when `res.headersSent` is already true.

This can happen for streaming responses or routes that call `res.write()` before `res.end()`. Previously, on-headers wrote cookies right before headers were sent, so this may be a behavior change where session updates are silently lost.

lib/client.js
```diff
@@ -17,35 +15,147 @@ function sortSpaceDelimitedString(string) {
...
+
+    // Add telemetry header if enabled
+    if (config.enableTelemetry) {
+      headers.set(
+        'Auth0-Client',
+        Buffer.from(JSON.stringify(telemetryHeader)).toString('base64'),
+      );
+    }
+
+    return fetch(fetchUrl, {
```
nandan-bhat: `httpAgent` is still listed as a public configuration option, but this new fetch implementation doesn’t seem to use it anymore. Some users might depend on it for things like proxies, custom TLS settings, or controlling network access.

Can we either make sure `httpAgent` is used in this new fetch logic, or clearly document this as a breaking change and update the types and docs?

package.json
```diff
@@ -34,9 +34,9 @@
     "futoin-hkdf": "^1.5.3",
     "http-errors": "^1.8.1",
     "joi": "^17.13.3",
-    "jose": "^2.0.7",
+    "jose": "^6.1.3",
     "on-headers": "^1.1.0",
```
nandan-bhat: Since this PR no longer uses `on-headers`, can we remove it from dependencies ?

If we decide to restore on-headers to preserve the old session-cookie timing behavior, then it should stay. Otherwise this looks like a stale dependency after the session persistence refactor.

lib/context.js
```diff
@@ -601,14 +755,77 @@ class ResponseContext {
...
+        headers.set(
+          'Auth0-Client',
+          Buffer.from(JSON.stringify(telemetryHeader)).toString('base64'),
+        );
+      }
+
+      // Use createRemoteJWKSet with proper caching, timeout, and custom fetch
+      // This provides automatic caching, respects SDK configuration, and handles key rotation
+      const jwksCache = oidcClient.getJwksCache(configuration);
+      const JWKS = createRemoteJWKSet(new URL(serverMetadata.jwks_uri), {
```
nandan-bhat: Can we confirm `JWKS` caching is effective across backchannel logout requests here ?

Since `createRemoteJWKSet()` is created inside each request handler call, each logout request appears to get a new remote JWKS resolver/cache.

If `oidcClient.getJwksCache(configuration)` returns undefined on a fresh process or before another flow has populated it, this may refetch JWKS more often than needed. Should we cache the `RemoteJWKSet` per issuer/configuration instead ?

V3_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,190 @@
...
+
+`v3.x` upgrades the underlying OpenID Connect and JWT dependencies (`openid-client` v4 → v6, `jose` v2 → v6) to their latest major versions, bringing improved security, performance, and standards compliance.
+
+**Important:** While this is a major version bump for the library, **there are ZERO breaking changes to the public API**. Your application code does not need to change.
```
nandan-bhat: Can we soften this claim or update it after the remaining compatibility questions are resolved ?

Right now we may still have user-visible behavior changes, like `options.redirectUri` no longer being honored and `httpAgent` no longer being used.

If those changes are intentional, they should be listed in this migration guide. If they are not intentional, we should fix them before saying there are zero public API breaking changes.

test/client.tests.js
```diff
@@ -208,39 +237,40 @@ describe('client initialization', function () {
...
+      const { configuration } = await getClient({
+        ...config,
+        httpTimeout: 1500,
+      });
+      const response = await invokeRequest(configuration);
+      assert.equal(response.status, 200);
     });
 
-    it('should timeout for delay > httpTimeout', async function () {
+    it.skip('should timeout for delay > httpTimeout', async function () {
```
nandan-bhat: Can we avoid skipping this test ? `httpTimeout` is still a public config option, so we should keep coverage that requests actually fail when they exceed the configured timeout.

If timeout behavior changed with `fetch/openid-client` v6, we should update the assertion to match the new error shape rather than skipping the test entirely.

test/client.tests.js
```diff
@@ -326,12 +369,13 @@ describe('client initialization', function () {
...
         .persist()
         .get('/.well-known/openid-configuration')
-        .reply(200, wellKnown);
+        .reply(200, { ...wellKnown, issuer: 'https://par-test.auth0.com/' });
       await expect(getClient(config)).to.be.fulfilled;
     });
   });
 
-  describe('client respects clientAssertionSigningAlg configuration', function () {
+  describe.skip('client respects clientAssertionSigningAlg configuration', function () {
```
nandan-bhat: Can we avoid skipping this test suite ? `clientAssertionSigningAlg` is still a public option, and this PR changes the private key JWT implementation during the `openid-client` v6 migration.

We should keep coverage that the configured signing algorithm is actually used, even if the test needs to be rewritten around the v6 grant APIs.

test/client.tests.js
```diff
@@ -266,17 +301,28 @@ describe('client initialization', function () {
...
+        'GET',
+      );
       expect(handler.firstCall.thisValue.req.headers['user-agent']).to.equal(
         'foo',
       );
     });
   });
 
-  describe('client respects httpAgent configuration', function () {
+  describe.skip('client respects httpAgent configuration', function () {
```
nandan-bhat: Can we avoid skipping this test suite ?. `httpAgent` is still a public config option, but this PR no longer appears to pass it into the HTTP layer.

If `httpAgent` support is intentionally removed because of the fetch migration, we should document it as a breaking change and update the public `types/docs`.

Otherwise, this test should be rewritten to verify the new implementation still honors `httpAgent`.

package-lock.json
```diff
@@ -48,10 +48,11 @@
         "sinon": "^7.5.0",
         "tsd": "^0.33.0",
         "typedoc": "^0.28.7",
-        "typescript": "^5.8.3"
+        "typescript": "^5.8.3",
+        "undici": "6.23.0"
       },
       "engines": {
-        "node": "^10.19.0 || >=12.0.0 < 13 || >=13.7.0 < 14 || >= 14.2.0"
+        "node": ">=20.0.0"
```
nandan-bhat: Can we keep this in sync with `package.json` ?

The package declares Node support as `^20.19.0 || ^22.12.0 || >= 23.0.0, but the lockfile root package still says >=20.0.0`. Take a look at this: https://github.com/auth0/express-openid-connect/blob/bfe775eb86982c21ccb4cd5098231b80c442280f/package.json#L76

Since the migration guide says the exact minimum versions matter for require(ESM) support, the lockfile should reflect the same engine range to avoid confusion.

index.d.ts
nandan-bhat: Refer to https://github.com/auth0/express-openid-connect/pull/785/changes#r3160619280

lib/context.js
cschetan77: yes, I completely agree, the support for `options.redirectUri` should not have been dropped. I already pointed this out earlier, I guess this got missed to address - https://github.com/auth0/express-openid-connect/pull/785#discussion_r3128074208
pre-v6, the `redirectUri` was constructed from either `options.redirectUri` or using `getRedirectUri` method and explicitly passed to openid-client's `callback` method. 
Have restored the same behavior.

cschetan77: Updated: 
1. Removed the comment that said `options.redirectUri` is no longer used
2. Added `callbackUrl` construction after `currentUrl` is built, restoring the `options.redirectUri || this.getRedirectUri()` fallback chain as earlier.
3. The `request` obj passed to oidcClient has this `callcackUrl` instead of original request's url

lib/appSession.js
cschetan77: Right, this is a limitation introduced by removing `on-headers`. It can all be attributed to jose-v6 async encryption, which makes our `setCookie` async too. `on-headers` only supports synchronous callbacks, so it can no longer be used to write the cookie right before headers are flushed. 

I've made the failure explicit with a debug log when `headersSent` is already true, covering not just streaming responses but any scenario where headers are flushed early (`res.write()`, `res.writeHead()`, `res.flushHeaders()`, `res.sendFile()`).

As per your suggestion I've also added a test that properly documents and asserts this behavior.

lib/client.js
cschetan77: yes, support for passing `httpAgent` has been removed. 

The core issue is that `httpAgent` worked because `got` (the HTTP library openid-client v4 used internally) had a first-class concept of agents, we could hand it an `http.Agent` or `https.Agent` and it would use that for every connection it made.
openid-client v6 replaced `got` with native `fetch`, and `fetch` simply has no equivalent concept, there's no way to pass a Node.js `http.Agent` into a fetch() call. The two APIs model transport-level control completely differently, so the old option can't be mapped across.

**Update**:  
Reference - [PR#787](https://github.com/auth0/express-openid-connect/pull/787/changes#diff-7aa4473ede4abd9ec099e87fec67fd57afafaf39e05d493ab4533acc38547eb8R737-R754)

Instead, we've exposed `customFetch` as the escape hatch. It gives consumers full control over the underlying HTTP behaviour, whether that's routing through a proxy, pinning a custom TLS certificate, adding retry logic, or just logging outbound requests for debugging. The SDK wraps whatever function is provided to inject required headers (User-Agent, telemetry) before the call goes out, so SDK concerns stay intact regardless of what the consumer does underneath. For the proxy case specifically, `undici` ships with Node.js 18+ so no extra dependency is needed

On the **type** side, removed `httpAgent` from `index.d.ts` (along with the HttpAgent/HttpsAgent imports) and typed `customFetch` as `typeof fetch`, This is broad enough to be compatible with `undici` and other fetch-compatible implementations. 

Also added Example-13 to `EXAMPLES.md` showing the undici ProxyAgent pattern end-to-end.

For test coverage, added two behavioural tests, one asserting the provided function is actually invoked during discovery, and another confirming the SDK's header decoration still runs before the call reaches the consumer's function.

package.json
cschetan77: Removed it from dep list. 
I know it was perfect at setting the cookie at right moment but don't think we can bring it back, as it expects a synchronous callback and our `setCookie` method changed to async.

lib/context.js
cschetan77: Yes, this was indeed creating a fresh `RemoteJWKSet` instance on every backchannel logout request, its internal key cache was thrown away after each verification.

The attempt to seed it via `oidcClient.getJwksCache(configuration)` was a partial workaround, but it only helped if openid-client had already fetched the JWKS as a side effect of a prior login or token refresh in the same process. On a fresh process, or when backchannel logout fires before any user has logged in (e.g. bulk session revocation at startup), `getJwksCache` returns `undefined` and every request goes cold to the IdP's JWKS endpoint anyway.

I agree, we should cache the `RemoteJWKSet` instance per config reference. 

**Update**: 
I've hoisted the `RemoteJWKSet` instance creation to module scope by defining `getRemoteJWKSet`, cached in a `jwksClientCache` Map keyed on the `config` reference, the same pattern `client.js` already uses for discovery caching. Since config is a stable object reference created once per `auth()` call and shared across all requests, this effectively gives us one `RemoteJWKSet` per SDK configuration for the lifetime of the process. 

Also, removed the the `oidcClient.getJwksCache()` seeding since it's now completely redundant.

test/client.tests.js
cschetan77: I've re enabled this test suite and updated the tests to cover - 
Test 1: should use RS256 as the default signing algorithm in the client assertion
Test 2: should use the configured signing algorithm in the client assertion
Test 3: should fail when signing algorithm is incompatible with the key type

test/client.tests.js
cschetan77: I've re enabled and updated the test suite `client respects httpTimeout configuration` to better cover for all 3 `httpTimeout` scenarios.

package-lock.json
cschetan77: Addressed it for now, 
But we will probably plan the removal of package lock file after after this in a separate PR.

V3_MIGRATION_GUIDE.md
cschetan77: Updated the migration guide with all the breaking changes, providing a summary at the start and then describing each breaking change in detail in their own section.

nandan-bhat (commented):
I'll add few more comments

nandan-bhat (dismissed):

lib/client.js
```diff
@@ -58,8 +179,8 @@ async function get(config) {
   const configRespType = sortSpaceDelimitedString(
     config.authorizationParams.response_type,
   );
-  const issuerRespTypes = Array.isArray(issuer.response_types_supported)
-    ? issuer.response_types_supported
+  const issuerRespTypes = Array.isArray(serverMetadata.response_types_supported)
+    ? serverMetadata.response_types_supported
     : [];
   issuerRespTypes.map(sortSpaceDelimitedString);
```
nandan-bhat: `.map()` returns a new array and does not change `issuerRespTypes`. Doesn't this mean the comparison on line `186` always uses the original unsorted values ? For example, wouldn't `"code id_token"` fail to match `"id_token code"` ?

Should this be:
```
const sortedRespTypes = issuerRespTypes.map(sortSpaceDelimitedString);
if (!sortedRespTypes.includes(configRespType)) {
```

index.d.ts
nandan-bhat, line 756: The TypeScript type still allows `'ES256K'` and `'EdDSA'`, but the Joi schema in https://github.com/auth0/express-openid-connect/blob/be90b7d936c0c6f005b4c6ee13e02cd91482c6ed/lib/config.js#L316-L328 rejects them at runtime.

Should we update the type to match the new valid set (`'Ed25519' `instead of `'EdDSA'`, and remove `'ES256K'`) ?

lib/client.js
cschetan77: yes, the comparison should have been made on `sortedRespTypes` result.
Updated.

index.d.ts
cschetan77: yes, there was a mismatch between Joi schema for `clientAssertionSigningAlg` and type def file. 
This was mistakenly left unchecked while determining breaking changes. 
Updated the `index.d.ts` to reflect the same.

lib/tokenset.js
```diff
@@ -0,0 +1,72 @@
...
+    this.id_token = tokenSet.id_token;
+    this.access_token = tokenSet.access_token;
+    this.refresh_token = tokenSet.refresh_token;
+    // Normalize token_type to 'Bearer' for backward compatibility
+    // openid-client v6 returns lowercase 'bearer'
+    this.token_type =
+      tokenSet.token_type?.toLowerCase() === 'bearer'
+        ? 'Bearer'
+        : tokenSet.token_type;
+
```
cschetan77: `token_type` normalization already happened during session write time. This is redundant.

nandan-bhat (approved):

cschetan77 (approved):

PR #812, opened 2026-04-05
fix: improve error message when transaction cookie state is missing

cschetan77:
### Description

Replaces the cryptic `checks.state argument is missing` error from `openid-client` with a descriptive message that identifies the affected cookie by name.


cschetan77:
Closing this PR, this fix will be implemented as part of #797

PR #828, opened 2026-06-11
feat: Respect IdP provided session expiry and cap the SDK session 

cschetan77:
### Description

When an upstream enterprise IdP (Okta, OIDC) emits a `session_expiry` claim in its ID token, Auth0 captures it and re-emits it in the ID token issued to the RP (when the connection has `id_token_session_expiry_supported: true`). Without SDK enforcement, Auth0 revokes the session server-side on the next `/authorize` round-trip, but the app-level session continues to look valid — defeating the security guarantee.

This PR makes the SDK enforce `session_expiry` as a hard ceiling on the local session lifetime.

**The change is non-breaking** — all enforcement is gated on the claim being present. Sessions without it (database/social logins, connection option disabled, or sessions created before this version) behave exactly as before.

---

### What it does

**At login (callback):**
- Extracts `session_expiry` from the ID token claims using strict positive-integer validation (fail-open: a missing or malformed claim means "no ceiling", never "already expired")
- Runs a lockout guard — rejects with HTTP 400 if `session_expiry` was already in the past when the IdP issued the token (compared against `claims.iat`, not `now`, so a token that arrives late is not incorrectly rejected); a born-dead session is never persisted
- Persists the ceiling as `session.sessionExpiresAt` (Unix seconds) in the session cookie

**On every request (session read in `appSession.js`):**
- Asserts `sessionExpiresAt` has not been reached (with 30s negative leeway for clock skew) alongside existing `iat`/`uat`/`exp` checks
- Gated on presence — sessions without `sessionExpiresAt` are unaffected

**On token refresh:**
- Pre-checks the ceiling before calling `/oauth/token` — throws `SessionExpiredError` instead of letting Auth0 return a confusing `invalid_grant`
- Preserves `sessionExpiresAt` across refreshes (the refresh grant does not re-assert `session_expiry`)

**Cookie `maxAge`:**
- `calculateExp()` now accepts an optional `sessionExpiresAt` and caps `exp = Math.min(exp, sessionExpiresAt)` — so the browser cookie never outlives the IdP session ceiling

---

### New public API

- **`SessionExpiredError`** — exported error class (`code: 'ERR_SESSION_EXPIRED'`, `status: 401`). Catch this in API routes to redirect for re-authentication rather than receiving a confusing `invalid_grant`.
- **`Session.sessionExpiresAt?: number`** — new optional session field (Unix seconds)
- **`IdTokenClaims.session_expiry?: number`** — new optional ID token claim type

---

### Files changed

| File | Change |
|---|---|
| `lib/errors.js` | New `SessionExpiredError` class |
| `lib/utils/epoch.js` | Extracted `epoch()` utility (shared by `appSession.js` and `context.js`) |
| `lib/utils/sessionExpiry.js` | New utility: `SESSION_EXPIRY_LEEWAY`, `isPlausibleUnixSeconds`, `extractSessionExpiry`, `isSessionExpiryReached`, `isSessionExpiryInPast` |
| `lib/context.js` | Callback: extract + validate + persist `sessionExpiresAt`; refresh: pre-check + preserve ceiling |
| `lib/appSession.js` | Session read: expiry assert; `calculateExp`: cap at `sessionExpiresAt` |
| `index.js` | Export `SessionExpiredError` |
| `index.d.ts` | Types for `SessionExpiredError`, `Session.sessionExpiresAt`, `IdTokenClaims.session_expiry` |

---

#### Unit tests

**`extractSessionExpiry` (`test/sessionExpiry.tests.js`)**
- Returns the value for a valid positive integer
- Returns `undefined` when the claim is absent (`undefined`, `null`, missing key)
- Returns `undefined` for a string value — fail-open
- Returns `undefined` for a float value — fail-open
- Returns `undefined` for zero — fail-open
- Returns `undefined` for a negative value — fail-open
- Returns `undefined` for `NaN` — fail-open
- Returns `undefined` for a millisecond-magnitude value — fail-open

**`isSessionExpiryReached` (`test/sessionExpiry.tests.js`)**
- Returns `false` when `sessionExpiresAt` is `undefined` (no ceiling)
- Returns `false` when ceiling is well in the future
- Returns `true` when ceiling is in the past
- Returns `true` when `now` is within the 30s leeway window (clock skew guard)
- Returns `true` exactly at the leeway boundary — inclusive
- Returns `false` just outside the leeway boundary

**`isSessionExpiryInPast` (`test/sessionExpiry.tests.js`)**
- Returns `false` when `sessionExpiresAt` is `undefined` (no ceiling)
- Returns `false` when ceiling is well in the future relative to `iat`
- Returns `true` when ceiling is in the past relative to `iat`
- Returns `true` when ceiling is within the 30s leeway window relative to `iat`
- Returns `true` exactly at the leeway boundary relative to `iat` — inclusive
- Returns `false` just outside the leeway boundary relative to `iat`
- Falls back to `epoch()` when `issuedAt` is absent
- Falls back to `epoch()` when `issuedAt` is a millisecond-magnitude value

**`appSession` session read (`test/appSession.tests.js`)**
- Session expires when `sessionExpiresAt` ceiling is reached
- Session remains valid when ceiling is in the future
- Sessions without `sessionExpiresAt` are unaffected — non-breaking
- Cookie `maxAge` is capped at `sessionExpiresAt` when it is sooner than `absoluteDuration`

**Callback + refresh (`test/callback.tests.js`)**
- `sessionExpiresAt` is persisted in the session when `session_expiry` claim is present
- `sessionExpiresAt` is absent when `session_expiry` claim is missing — non-breaking
- Login rejected with 400 when `session_expiry` is already in the past (lockout guard)
- Invalid `session_expiry` shapes (string, float, zero, negative) are ignored — fail-open
- `SessionExpiredError` (401) is thrown on `accessToken.refresh()` when ceiling has passed — `/oauth/token` is never called
- `sessionExpiresAt` is preserved in the session after a successful refresh (carry-forward)

#### Manual tests

Testing against a live Auth0 tenant with a Post-Login Action deployed to emit `session_expiry`.

- Login with `session_expiry` in ID token — `sessionExpiresAt` persisted in session
- Session read before expiry — session valid and accessible 
- Token refresh before expiry — new access token returned successfully 
- `sessionExpiresAt` preserved after refresh — ceiling carried forward across token refresh 
- Session read after expiry — session cleared, redirected to Auth0 for re-authentication 
- Token refresh after expiry — `SessionExpiredError` (401, `ERR_SESSION_EXPIRED`) thrown; `/oauth/token` not called


lib/utils/sessionExpiry.js
```diff
@@ -0,0 +1,32 @@
...
+const SESSION_EXPIRY_LEEWAY = 30;
+
+/**
+ * Reads the IPSIE `session_expiry` claim from ID token claims.
+ * Fail-open: returns undefined unless the value is a positive integer.
+ * A missing or malformed claim must never be treated as already-expired.
+ */
+function extractSessionExpiry(claims) {
+  const value = claims?.session_expiry;
+  return typeof value === 'number' && Number.isInteger(value) && value > 0
```
kishore7snehil: Since this claim is set by a customer-side action rather than the platform, it's effectively untrusted input- and the easiest mistake someone can make is handing it milliseconds instead of seconds (forgetting the `/1000`). A 13-digit millisecond value sails through this positive-integer check, lands tens of thousands of years in the future, and the ceiling then never triggers, so enforcement is silently off while everything *looks* fine.

Could we add an upper-bound sanity check here and treat anything implausibly large as "no ceiling" (fail-open)? Something like rejecting values `>= 10_000_000_000` , that's around the year 2286 in seconds, so no real timestamp is affected for a couple of centuries, but every millisecond slip-up gets caught.

lib/context.js
```diff
@@ -736,14 +758,31 @@ class ResponseContext {
...
       session.sid = claims?.sid;
 
+      // Persist the session_expiry claim from the ID token as sessionExpiresAt.
+      // This is the IdP-asserted ceiling on the RP session lifetime (IPSIE SL1).
+      // Only set when the claim is present — absence means the connection does not
+      // have id_token_session_expiry_supported enabled, and existing behavior applies.
+      const sessionExpiresAt = extractSessionExpiry(claims);
+      if (sessionExpiresAt) {
+        // Lockout guard: reject a session already past its ceiling at login.
+        if (sessionExpiresAt <= epoch()) {
```
kishore7snehil: The login guard here compares against `epoch()` with no leeway, but the read-time and refresh checks go through the helper that applies the 30s skew leeway. The effect is that a ceiling landing 1–30s in the future passes login and gets persisted, then the very next session read immediately rejects it ,so the user "logs in" and is bounced on the next page load, which is the born-dead session this guard is meant to avoid. 

Routing this through the same helper the reads use would keep both ends in agreement (and a quick boundary test would lock it in).

lib/context.js
```diff
@@ -143,6 +149,16 @@ function normalizeTokenType(tokenType) {
 
 async function refresh({ tokenEndpointParams } = {}) {
   let { config, req } = weakRef(this);
+  const session = req[config.session.name];
+
+  // Pre-check: do not call /oauth/token if the IdP session ceiling has passed.
+  // Auth0 has already revoked the session server-side at this point, so a
+  // refresh attempt would fail with invalid_grant anyway. Surfacing a clean
+  // SessionExpiredError is better than a confusing token-endpoint error.
+  if (isSessionExpiryReached(session.sessionExpiresAt)) {
```
kishore7snehil: `session` is read and de-referenced right at the top, so if `refresh()` is ever reached without a live session this throws a `TypeError` before the usual token checks can surface a clearer error.

lib/errors.js
```diff
@@ -0,0 +1,11 @@
+class SessionExpiredError extends Error {
+  constructor(message = 'The upstream IdP session has expired.') {
+    super(message);
+    this.name = 'SessionExpiredError';
+    this.code = 'ERR_SESSION_EXPIRED';
```
kishore7snehil, line 5: This is a new exported error, but right now it only ever surfaces on `refresh()`. The read path silently clears the session (no error reaches the app) and the login lockout throws a generic 400 instead ,so the three places a ceiling can be hit each signal differently. 

Do we want this to be the one consistent "session expired" signal across all three sites, or is it deliberately refresh-only?

.npmrc
```diff
@@ -1,3 +0,0 @@
-registry=https://registry.npmjs.org/
```
kishore7snehil, line 1: This looks unrelated to the feature — deleting `.npmrc` drops `save-exact=true` (plus the registry/loglevel pins), which means new dependencies start getting saved as caret ranges instead of exact versions and installs become less reproducible. Was this intentional? If not, probably worth restoring; if it is, a line in the description would help.

test/sessionExpiry.tests.js
```diff
@@ -0,0 +1,75 @@
...
+
+  it('returns undefined for zero (fail-open)', () => {
+    assert.isUndefined(extractSessionExpiry({ session_expiry: 0 }));
+  });
+
+  it('returns undefined for a negative value (fail-open)', () => {
+    assert.isUndefined(extractSessionExpiry({ session_expiry: -100 }));
+  });
+
+  it('returns undefined for NaN (fail-open)', () => {
```
kishore7snehil, line 39: The one gap I'd add (pairs with the upper-bound suggestion on the helper): a test for an implausibly large, millisecond-magnitude value, asserting it's treated as "no ceiling" rather than a far-future one.

test/callback.tests.js
```diff
@@ -1440,4 +1440,226 @@ describe('callback response_mode: form_post', () => {
...
+        authorizationParams: {
+          response_type: 'code id_token',
+          scope: 'openid profile email offline_access',
+        },
+      };
+      const router = auth(authOpts);
+
+      // This route bypasses appSession expiry check by directly manipulating the
+      // session ceiling to simulate it having passed, then calling refresh().
+      router.get('/refresh-expired', async (req, res, next) => {
```
kishore7snehil, line 1587: This proves the refresh pre-check in isolation but the `/refresh-expired` backdoor sets the ceiling directly and skips the middleware read-check, so the realistic end-to-end path isn't exercised. 

Could we add a fake-timer test that advances past the ceiling and lets the middleware clear the session naturally, plus a boundary case (ceiling ~15s out) that documents what should happen at login vs. the next read? That second one would also catch the leeway mismatch flagged in context.js.

kishore7snehil (changes requested):

lib/errors.js
cschetan77: No, this was intentional. 

- Session read silently clearing is correct behavior. The middleware should not throw, it should just treat the session as gone and let the `auth` middleware redirect to login as it would for any unauthenticated request. Throwing an error here would break every app that doesn't have explicit error handling.
- Login lockout is a different failure mode entirely, it's a bad token at login time, not an expired session. A `400` is appropriate, and SessionExpiredError (401) would be semantically wrong there.

`refresh()` is the only place where app code explicitly calls an async operation and needs a catchable, typed error to decide what to do (redirect vs. retry vs. surface to user).

.npmrc
cschetan77: This was accidental, have restored the file.
Though was planning to remove registry url from it, but will do that in a  separate PR.

test/callback.tests.js
cschetan77: Added two new tests to cover:
- A fake-timer test that advances the clock past the ceiling and lets `appSession.js` naturally clear the session on the next request, no backdoor involved.
- A leeway boundary test that documents what happens at login when `session_expiry` lands within the 30s leeway window (rejected with 400) vs. just outside it (login succeeds), this also locks in the fix for the leeway mismatch flagged in `context.js`.

lib/utils/sessionExpiry.js
cschetan77: Yes, that makes sense and the value `10_000_000_000` seems the best choice. 
Added an upper bound of `10_000_000_000` (year 2286) to extractSessionExpiry, any value at or above that, is treated as "no ceiling" (fail-open), which catches every accidental millisecond slip-up while leaving all realistic seconds timestamps unaffected for the next couple of centuries. 
Also added a test asserting a millisecond-magnitude value is ignored.

lib/context.js
cschetan77: Updated: The lockout guard now routes through `isSessionExpiryReached` instead of comparing against epoch() directly.

lib/context.js
cschetan77: Changed to `session?.sessionExpiresAt`, if session is null or undefined, `isSessionExpiryReached(undefined)` returns false, and the code falls through to fail naturally at the token endpoint call rather than throwing a misleading `SessionExpiredError`.

test/sessionExpiry.tests.js
cschetan77: Added a test asserting that a millisecond-magnitude value (1748566800000) is treated as "no ceiling", pairs with the upper-bound fix on `extractSessionExpiry`.

lib/errors.js
```diff
@@ -0,0 +1,11 @@
+class SessionExpiredError extends Error {
```
kishore7snehil, line 1: This adds `SessionExpiredError` as a public export and a new `sessionExpiresAt` field on the session, both well-described in the typings but there's no prose documentation for the feature anywhere in the repo. EXAMPLES.md, README.md, and FAQ.md don't mention session_expiry, and the diff doesn't touch any markdown.

You can refer to this - https://github.com/auth0/auth0-auth-js/pull/191/changes#diff-15a5f387b45be22cd46b0b8293094d414345ca978c918743c43e0a9537964d65

kishore7snehil (changes requested):

lib/errors.js
cschetan77: Added documentation across three files:

- `EXAMPLES.md`: `"Session expiry from upstream IdP (IPSIE session_expiry)"` covering what it's about and what's the SDK does automatically (lockout guard, session enforcement, refresh pre-check, cookie cap), behavior on expiry with a `SessionExpiredError` catch example, and how to read `sessionExpiresAt`.
- `FAQ.md`: new entry for the most likely confusing symptom, a previously authenticated user being unexpectedly redirected to login, pointing to the `EXAMPLES.md` section for full details.
- `README.md`: added a link to `EXAMPLES.md` in the Documentation section, which was missing for quick navigation.

EXAMPLES.md
```diff
@@ -432,3 +433,63 @@ app.use(
...
+
+## 14. Session expiry from upstream IdP (IPSIE `session_expiry`)
+
+When an upstream IdP supports the IPSIE SL1 spec, it can include a `session_expiry` claim in the ID token — an absolute Unix timestamp (seconds) marking the latest moment the IdP considers the session valid. Auth0, for example, emits this claim on enterprise connections configured with `id_token_session_expiry_supported: true`.
+
+### What the SDK does automatically
+
+No configuration or code change is required. When the claim is present, the SDK handles everything automatically:
+
+- Persists the ceiling as `sessionExpiresAt` (Unix seconds) on the session.
```
kishore7snehil, line 452: There's no mention of the milliseconds footgun or the fail-open behavior, even though that's exactly what the `10_000_000_000` guard added in this PR protects against.

[auth0-server-js EXAMPLES.md](https://github.com/auth0/auth0-auth-js/blob/cef72915050abe34a921f8fa1566774c76f94511/packages/auth0-server-js/EXAMPLES.md#session-expiry-from-upstream-idp-ipsie-session_expiry) and [auth0-server-python RetrievingData.md](https://github.com/auth0/auth0-server-python/blob/f29e406/examples/RetrievingData.md#session-expiry-from-the-upstream-idp) - server-python gives it a dedicated warning ("a milliseconds value will silently disable enforcement… always divide by 1000").

Could we add a short warning noting the claim must be seconds, that values >= 10,000,000,000 are treated as "no ceiling," and that any malformed value fails open?

EXAMPLES.md
```diff
@@ -432,3 +433,63 @@ app.use(
...
+- Rejects login with HTTP 400 if the ceiling is already in the past at callback time, so a born-dead session is never persisted.
+- Treats the session as expired once `sessionExpiresAt` is reached on every request, with a 30-second leeway for clock skew.
+- Throws `SessionExpiredError` on `accessToken.refresh()` instead of making a token endpoint call that would fail anyway.
+- Caps the session cookie lifetime at the ceiling as a defense-in-depth backstop.
+
+This is layered **on top of** your existing idle and absolute session timeouts — the session ends at whichever limit is reached first.
```
kishore7snehil, line 458: Smaller thing, to match the depth of the [docs](https://github.com/auth0/auth0-auth-js/blob/cef72915050abe34a921f8fa1566774c76f94511/packages/auth0-server-js/EXAMPLES.md#session-expiry-from-upstream-idp-ipsie-session_expiry): they explain where the value comes from - Auth0 computes it as the earliest of the tenant's absolute session lifetime, the upstream IdP's expiry, and any `api.session.setExpiresAt` override, on okta/oidc enterprise connections.

kishore7snehil (changes requested):

EXAMPLES.md
```diff
@@ -432,3 +433,63 @@ app.use(
 ```
 
 The SDK wraps your `customFetch` function to add required headers (User-Agent, Auth0-Client telemetry) before making requests.
+
+## 14. Session expiry from upstream IdP (IPSIE `session_expiry`)
+
+When an upstream IdP supports the IPSIE SL1 spec, it can include a `session_expiry` claim in the ID token — an absolute Unix timestamp (seconds) marking the latest moment the IdP considers the session valid. Auth0, for example, emits this claim on enterprise connections configured with `id_token_session_expiry_supported: true`.
```
kishore7snehil: server-python handles this with a short "Emitting the claim" note plus a placeholder link to the canonical Auth0 Action guide: [RetrievingData.md](https://github.com/auth0/auth0-server-python/blob/f29e406/examples/RetrievingData.md#session-expiry-from-the-upstream-idp) ("For the canonical Action setup, see the Auth0 documentation (will be adding the link to the session_expiry Action guide once published)"). Could we add the same here - a line on the Action emitting the claim, with a placeholder for the Auth0 docs link to fill in once that guide is published?

kishore7snehil (changes requested):

kishore7snehil (dismissed):
LGTM!

lib/context.js
```diff
@@ -736,14 +758,31 @@ class ResponseContext {
...
       session.sid = claims?.sid;
 
+      // Persist the session_expiry claim from the ID token as sessionExpiresAt.
+      // This is the IdP-asserted ceiling on the RP session lifetime (IPSIE SL1).
+      // Only set when the claim is present — absence means the connection does not
+      // have id_token_session_expiry_supported enabled, and existing behavior applies.
+      const sessionExpiresAt = extractSessionExpiry(claims);
+      if (sessionExpiresAt) {
+        // Lockout guard: reject a session already past its ceiling at login.
+        if (isSessionExpiryReached(sessionExpiresAt)) {
```
nandan-bhat: This lockout compares `session_expiry` against `now`, but at login I think it should compare against the token's `iat` instead.

The point of this guard is to reject a session that is born expired, where `session_expiry <= iat`. Using `now` means a token whose ceiling is, say, 20s in the future passes here at login, but then fails the very next session read and the user gets bounced immediately. Comparing against `iat` avoids that confusing "logged in then instantly logged out" case.

Related point: `isSessionExpiryReached` is now doing double duty. The read path uses it for "has the ceiling passed `now` ?" and the login path reuses it for "was this born expired ?". Those are two different questions. `auth0-server-js` keeps them separate with a dedicated `iat` based helper (`isSessionExpiryInPast(sessionExpiresAt, claims.iat)`). Could we align with that here ? Keeping the two SDKs consistent would help.

lib/context.js
```diff
@@ -736,14 +758,31 @@ class ResponseContext {
...
+          throw createError(
+            400,
+            'The session_expiry claim is in the past at login time.',
+            { error: 'invalid_session_expiry' },
```
nandan-bhat, line 783: The login lockout throws a generic `createError(400, ...)` with `error: 'invalid_session_expiry'`, but the refresh path throws the new `SessionExpiredError` (401). Same underlying reason, "session is past its ceiling", surfaced as two different error types and status codes.

A consumer catching `SessionExpiredError` won't catch this login-time case. Could we throw `SessionExpiredError` here too, so there's just one thing to catch ? If the `400` is intentional (born-dead at login being a bad request rather than unauthorized), a short comment explaining the choice would help.

index.d.ts
```diff
@@ -36,9 +36,16 @@ interface AuthorizationParameters {
...
 /**
- * ID Token claims
+ * ID Token claims, extended with the IPSIE SL1 session_expiry claim.
  */
-type IdTokenClaims = IDToken;
+type IdTokenClaims = IDToken & {
+  /**
+   * Unix timestamp (seconds since epoch) indicating the hard ceiling on the RP
+   * session lifetime, as asserted by the upstream enterprise IdP.
+   * Present only when the connection has `id_token_session_expiry_supported: true`.
```
nandan-bhat: This SDK is a generic OIDC library, not Auth0-specific, but the docs and types here read as Auth0-only. `id_token_session_expiry_supported` is an Auth0 connection flag, and `session_expiry` is actually an IPSIE / OIDC profile claim that any conforming provider can emit.

Could we frame this generically, something like "when the OpenID Provider emits a `session_expiry` claim (per IPSIE SL1 §3.2.2)", and mention the Auth0 connection flag only as one example of how a provider enables it ? Otherwise a non-Auth0 user might assume this feature doesn't apply to them. Same applies to the "upstream enterprise IdP" wording just above and the `res.oidc.login()` example in the `SessionExpiredError` docblock.

lib/appSession.js
```diff
@@ -101,22 +102,36 @@ module.exports = (config) => {
     throw lastError;
   }
 
-  function calculateExp(iat, uat) {
+  // Calculates the session expiry Unix timestamp (seconds).
+  // iat: session issued-at time (Unix seconds)
+  // uat: session last-updated-at time (Unix seconds)
+  // sessionExpiresAt: optional IdP-asserted ceiling (Unix seconds) from IPSIE session_expiry claim;
+  //                   when present, caps the result so the cookie never outlives the IdP session.
+  function calculateExp(iat, uat, sessionExpiresAt) {
```
nandan-bhat, line 110: The cookie `maxAge` cap looks right, `exp = Math.min(exp, sessionExpiresAt)`. Do we have a unit test that asserts the cap actually kicks in when `sessionExpiresAt` is sooner than the rolling / absolute duration ? I saw the read-enforcement and callback tests, just want to make sure the cookie-cap path has an explicit assertion so it doesn't silently regress.

lib/context.js
cschetan77: This was an intentional decision, [discussion_r3451378840](https://github.com/auth0/express-openid-connect/pull/828#discussion_r3451378840).
- SessionExpiredError (401) is designed for the refresh path: the user had a valid, established session that has since expired. It's meant to be caught by app error handlers to trigger re-authentication.
- At the callback, the user has no established session yet, the IdP issued a token with a session_expiry already in the past. That's an invalid token, not an expired session. A 400 correctly signals "bad auth response."

index.d.ts
cschetan77: yes, that's appropriate. 
Updated both docstrings:
- `IdTokenClaims.session_expiry`: now reads "asserted by the upstream IdP via the IPSIE SL1 protocol", with id_token_session_expiry_supported mentioned as an Auth0-specific example.
- `Session.sessionExpiresAt`: same wording, generic framing first with Auth0 as the example.

`SessionExpiredError` jsdoc was already generic so no change needed there.

lib/appSession.js
cschetan77: The test was already there but only asserted `isBelow(maxAgeSeconds, 3 * 60 * 60)`, proving the cap fired but not that it landed specifically at sessionExpiresAt. Tightened it to:
```js
  const expSeconds = Math.floor(expDate.getTime() / 1000);
  assert.approximately(expSeconds, sessionExpiresAt, 5);
```
This confirms the cookie expiry is pinned to sessionExpiresAt

lib/context.js
cschetan77: yes, that makes sense. 
Updated the lockout guard to compare against `claims.iat` instead of now. 
- Added `isSessionExpiryInPast(sessionExpiresAt, issuedAt)` to `sessionExpiry.js` utility,  it resolves the reference time from iat (with a plausibility guard that falls back to epoch() if iat is absent or a millisecond-magnitude value), then delegates to `isSessionExpiryReached` so the leeway logic isn't duplicated. Also extracted `isPlausibleUnixSeconds` as a shared private helper used by both extractSessionExpiry and isSessionExpiryInPast.
This aligns with `auth0-server-js`.

nandan-bhat (approved):

PR #838, opened 2026-06-26
fix: restore on-headers for session cookie to support streaming responses

### LINKED ISSUE #835: Session cookie silently dropped when response headers are flushed before res.end()
## Problem

In v3, session cookie persistence moved from an `on-headers` hook (which fires at `res.writeHead()` time) to a `res.end()` wrapper. This means any response that flushes headers before `res.end()` — via `res.flushHeaders()`, `res.write()`, `res.writeHead()`, `res.sendFile()`, or `res.download()` — will have `res.headersSent === true` by the time the cookie write runs, and the session cookie is **silently dropped**.

This affects any application using streaming/chunked responses, which is a common pattern for server-side rendered applications that flush the `<head>` early to allow the browser to start loading assets while the body renders.

There is no error, no warning, and no workaround. The session just stops being persisted.

## Expected behavior

The session cookie should be written regardless of whether the response is streamed or buffered. The v2 behavior (using `on-headers`) handled this correctly.

## Reproduction

```js
app.use(auth(config));

app.get('/', (req, res) => {
  res.flushHeaders();       // headers sent
  res.write('<html>...');   // streaming body
  // ... later ...
  res.end('</html>');       // session cookie write runs here, but headersSent is true — cookie lost
});
```

## Suggested fix

Expose a mechanism to write the session cookie before the response starts streaming. Either:

1. **Restore `on-headers` as the default** — this was the v2 behavior, it fires at `writeHead()` time (the last moment headers are mutable), and it works with both buffered and streamed responses.

2. **Expose a public API** like `req.oidc.persistSession(res)` that consumers can call explicitly when they know headers are about to flush. This would allow streaming applications to call it at the right moment without relying on internal timing.

Option 1 is backward-compatible and covers all cases transparently. Option 2 gives consumers control but requires awareness of the issue.

## Environment

- `express-openid-connect`: v3 (breaking change from v2)
- Node.js: >=22
- Express: 4.x
- Use case: streaming SSR with `res.flushHeaders()` before `res.end()`

cschetan77:
## Summary

Fixes the silent session cookie loss introduced in v3 for any response that flushes headers before `res.end()` — including streaming SSR, `res.flushHeaders()`, `res.write()`, `res.sendFile()`, and `stream.pipe(res)`.

Closes #835

## Problem

v3 moved session cookie persistence from an `on-headers` hook to a `res.end()` wrapper. This was necessary because `jose` v6 moved to an async-only encryption API, and `on-headers` requires a synchronous callback. The consequence was that any response that flushes headers early (a standard SSR performance pattern) silently dropped the session cookie with no error or warning.

## Solution

Restore `on-headers` as the mechanism for writing the session cookie, and introduce two synchronous crypto primitives (using Node's built-in `crypto` module) to work around the async constraint:

- **`encryptSync`** — synchronous AES-256-GCM JWE (`alg=dir`, `enc=A256GCM`) in Compact Serialization format. Produces output byte-for-byte compatible with `jose` v6 `compactDecrypt` — existing sessions decrypt transparently with no migration needed.
- **`signCookieSync`** — synchronous HMAC-SHA256 JWS signature for signed custom store session ID cookies. Compatible with the existing `verifyCookie` verification path.

Both are standard cryptographic primitives — no novel crypto, just calling `crypto.createCipheriv` and `crypto.createHmac` directly, which are the same underlying operations `jose` performs internally.

## Behaviour restored

- Cookie-based sessions: `on-headers` → `encryptSync` → `res.setHeader('Set-Cookie', ...)`
- Custom store sessions: `on-headers` → `signCookieSync` (if signing enabled) → `res.setHeader('Set-Cookie', ...)`, plus `res.end` wrapper for the async `store.set()` persistence
- Streaming responses (`res.write()`, `res.flushHeaders()`, `res.sendFile()`, `stream.pipe(res)`): session cookie is now correctly set in all cases

## Testing

- 341 tests passing (19 new unit tests for `encryptSync` and `signCookieSync` in `test/crypto.tests.js`)
- Updated `test/appSession.tests.js` to assert the fixed behaviour (cookie IS written on streaming responses)
- Added streaming test to `test/appSession.customStore.tests.js` — verifies `Set-Cookie` is present when `res.write()` is called before `res.end()` on the custom store path
- Added test documenting that `Set-Cookie` is sent even when `store.set()` fails — consistent with v2 behaviour (cookie written at `writeHead` time, before persistence)
- Cross-compatibility verified: `encryptSync` output decrypted by `jose` v6 `compactDecrypt` ✓, `signCookieSync` output verified by existing `verifyCookie` 

## Test plan

- All existing tests pass
- New `crypto.tests.js` suite passes (`encryptSync` + `signCookieSync`)
- Streaming test passes for both cookie store and custom store paths
- Manual verification: streaming SSR route sets session cookie correctly after `res.write()` / `res.flushHeaders()`
- Existing sessions (encrypted by jose v6) are still readable after upgrade

## Smoke Testing

Tested manually against a real Auth0 tenant.

**CookieStore (default)**
- Login → session established, JWE cookie set correctly
- Session readable across requests (decrypt round-trip verified)
- Session mutation persists on subsequent requests
- Logout → session cleared, cookie absent

**CustomStore (signed session cookie)**
- Login → short session ID + HMAC signature written to cookie, session data persisted to store
- Session readable across requests (store lookup + signature verification)
- Session mutation → store updated, cookie re-signed
- Logout → store entry destroyed, cookie cleared

**Streaming fix (core regression — both store modes)**
- `res.write()` before `res.end()` → `Set-Cookie: appSession=` present in response headers 
- `res.flushHeaders()` before write → `Set-Cookie: appSession=` present 
- `stream.pipe(res)` → `Set-Cookie: appSession=` present 

All three streaming scenarios failed silently (no `Set-Cookie`) in v3.0.0 before this fix.

### DIFF
--- a/lib/appSession.js
+++ b/lib/appSession.js
@@ -1,13 +1,18 @@
 const { strict: assert, AssertionError } = require('assert');
 const {
-  CompactEncrypt,
   compactDecrypt,
   errors: { JOSEError },
 } = require('jose');
+const onHeaders = require('on-headers');
 const safePromisify = require('./utils/promisifyCompat');
 const cookie = require('cookie');
 const COOKIES = require('./cookies');
-const { getKeyStore, verifyCookie, signCookie } = require('./crypto');
+const {
+  getKeyStore,
+  verifyCookie,
+  signCookieSync,
+  encryptSync,
+} = require('./crypto');
 const debug = require('./debug')('appSession');
 
 const { epoch } = require('./utils/epoch');
@@ -73,14 +78,6 @@ module.exports = (config) => {
   let [current, keystore] = getKeyStore(config.secret, true);
   // keystore is now an array of keys or a single key buffer
 
-  async function encrypt(payload, headers) {
-    const encoder = new TextEncoder();
-    const protectedHeader = { alg, enc, ...headers };
-    return await new CompactEncrypt(encoder.encode(payload))
-      .setProtectedHeader(protectedHeader)
-      .encrypt(current);
-  }
-
   async function decrypt(jwe) {
     let lastError;
 
@@ -124,7 +121,7 @@ module.exports = (config) => {
     return exp;
   }
 
-  async function setCookie(
+  function setCookie(
     req,
     res,
     {
@@ -133,13 +130,6 @@ module.exports = (config) => {
       exp = calculateExp(iat, uat, req[sessionName]?.sessionExpiresAt),
     },
   ) {
-    if (res.headersSent) {
-      // If headers were already flushed before res.end() the
-      // session cookie will not be persisted for this response.
-      debug('session cookie could not be set: headers already sent');
-      return;
-    }
-
     const cookies = req[COOKIES];
     const { transient: cookieTransient, ...cookieOptions } = cookieConfig;
     cookieOptions.expires = cookieTransient ? 0 : new Date(exp * 1000);
@@ -161,7 +151,7 @@ module.exports = (config) => {
         sessionName,
       );
 
-      const value = await encrypt(JSON.stringify(req[sessionName]), {
+      const value = encryptSync(JSON.stringify(req[sessionName]), current, {
         iat,
         uat,
         exp,
@@ -219,8 +209,8 @@ module.exports = (config) => {
       return req[COOKIES][sessionName];
     }
 
-    async setCookie(req, res, iat) {
-      await setCookie(req, res, iat);
+    setCookie(req, res, iat) {
+      setCookie(req, res, iat);
     }
   }
 
@@ -284,7 +274,7 @@ module.exports = (config) => {
       return req[COOKIES][sessionName];
     }
 
-    async setCookie(
+    setCookie(
       id,
       req,
       res,
@@ -294,14 +284,8 @@ module.exports = (config) => {
         exp = calculateExp(iat, uat, req[sessionName]?.sessionExpiresAt),
       },
     ) {
-      if (res.headersSent) {
-        // If headers were already flushed before res.end() the
-        // session cookie will not be persisted for this response.
-        debug('session cookie could not be set: headers already sent');
-        return;
-      }
       if (!req[sessionName] || !Object.keys(req[sessionName]).length) {
-        if (req[COOKIES][sessionName] && !res.headersSent) {
+        if (req[COOKIES][sessionName]) {
           clearCookie(sessionName, res);
         }
       } else {
@@ -312,12 +296,9 @@ module.exports = (config) => {
         delete cookieOptions.transient;
         let value = id;
         if (signSessionStoreCookie) {
-          value = await signCookie(sessionName, id, this._current);
+          value = signCookieSync(sessionName, id, this._current);
         }
-        // Check again after async operation in case headers were sent
-        if (!res.headersSent) {
-          res.cookie(sessionName, value, cookieOptions);
-        }
+        res.cookie(sessionName, value, cookieOptions);
       }
     }
   }
@@ -449,19 +430,19 @@ module.exports = (config) => {
       attachSessionObject(req, sessionName, {});
     }
 
-    const { end: origEnd } = res;
-
     if (isCustomStore) {
       const id = existingSessionValue || (await generateId(req));
 
+      onHeaders(res, () =>
+        store.setCookie(req[REGENERATED_SESSION_ID] || id, req, res, { iat }),
+      );
+
+      const { end: origEnd } = res;
       res.end = async function resEnd(...args) {
         try {
           // Persist session data first, then set cookie only if persistence succeeds.
           // This prevents sending a cookie pointing to data that was never persisted.
           await store.set(id, req, res, {
-            iat,
-          });
-          await store.setCookie(req[REGENERATED_SESSION_ID] || id, req, res, {
             iat,
           });
           origEnd.call(res, ...args);
@@ -473,15 +454,7 @@ module.exports = (config) => {
         }
       };
     } else {
-      res.end = async function resEnd(...args) {
-        try {
-          await store.setCookie(req, res, { iat });
-          origEnd.call(res, ...args);
-        } catch (e) {
-          res.end = origEnd;
-          process.nextTick(() => next(e));
-        }
-      };
+      onHeaders(res, () => store.setCookie(req, res, { iat }));
     }
 
     return next();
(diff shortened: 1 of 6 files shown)


lib/appSession.js
```diff
@@ -416,21 +397,21 @@ module.exports = (config) => {
       attachSessionObject(req, sessionName, {});
     }
 
-    const { end: origEnd } = res;
-
     if (isCustomStore) {
       const id = existingSessionValue || (await generateId(req));
 
+      onHeaders(res, () =>
+        store.setCookie(req[REGENERATED_SESSION_ID] || id, req, res, { iat }),
```
nandan-bhat, line 437: One concern here. In v3.0.0 the custom store cookie was written inside the `res.end` wrapper, right after `store.set()` succeeded, and the comment there said we only set the cookie if persistence succeeds. Now that `setCookie` moved into `onHeaders`, it fires at `writeHead` time, which is before `store.set()` runs in `res.end`. So if `store.set()` fails against Redis, the browser still gets a `Set-Cookie` pointing to a session id that was never saved, and on the next request the user looks logged out. Can we gate this cookie write on `store.set()` succeeding, or keep it in `res.end` for the custom store path?

lib/appSession.js
```diff
@@ -416,21 +397,21 @@ module.exports = (config) => {
...
+      );
+
+      const { end: origEnd } = res;
       res.end = async function resEnd(...args) {
         try {
           // Persist session data first, then set cookie only if persistence succeeds.
```
nandan-bhat: Small thing: this comment says we persist first and then set the cookie only if persistence succeeds, but `store.setCookie` is no longer called in this block. It now runs in the `onHeaders` callback above, so the comment is misleading now. Can we update or remove it?

lib/appSession.js
```diff
@@ -440,15 +421,7 @@ module.exports = (config) => {
...
-      res.end = async function resEnd(...args) {
-        try {
-          await store.setCookie(req, res, { iat });
-          origEnd.call(res, ...args);
-        } catch (e) {
-          res.end = origEnd;
-          process.nextTick(() => next(e));
-        }
-      };
+      onHeaders(res, () => store.setCookie(req, res, { iat }));
```
nandan-bhat, line 455: In v3.0.0 this cookie write was wrapped in a `try/catch` and any error was passed to `next(e)`. Now if `encryptSync` throws inside the `onHeaders` callback, the error is not caught and would surface at `writeHead` time instead of going through the Express error handler. It is a rare case with valid keys, but it is a behavior change from v3.0.0. Worth a `try/catch` here to keep it consistent.

test/appSession.tests.js
```diff
@@ -465,27 +465,25 @@ describe('appSession', () => {
     clock.restore();
   });
 
-  it('should not write session cookie when headers are already sent before res.end()', async () => {
+  it('should write session cookie even when headers are flushed before res.end()', async () => {
```
nandan-bhat, line 469: This updated test covers the cookie store path. Could we also add a test for the custom store path with `onHeaders`, and one where `store.set()` throws, to confirm we do not send a `Set-Cookie` for a session that was never persisted? That is the case I am most worried about with the new ordering.

lib/appSession.js
cschetan77: This ordering (cookie written at `writeHead` time, `store.set()` running after in `res.end`) is actually identical to how v2 behaved. In v2, `on-headers` fired the cookie write and `store.set()` ran inside the res.end wrapper afterwards, so the browser could always receive a Set-Cookie pointing to a session ID that hadn't been persisted yet, in the event of a store failure.

`v3.0.0` did genuinely improve on this by moving `setCookie` inside `res.end` after `store.set()`, gating the cookie on successful persistence. However, that improvement is what caused the streaming regression, the move to `res.end()` was necessary because of `jose` v6's async-only encryption API, but as a side effect it broke any response that flushes headers before res.end().

There is no way to both write the cookie before headers flush (required for streaming) and guarantee the store has already succeeded (requires waiting until after res.end). These two guarantees are mutually exclusive.

Given that the v2 ordering shipped without issues being reported against it, and that the streaming regression is actively breaking production applications, we should restore the v2 behaviour and that seems the right trade-off. 
The cookie-after-persist guarantee was a v3.0.0 improvement that no one had before, so this is not a regression against any previously shipped behaviour.

lib/appSession.js
cschetan77: You are correct, that it's a behaviour change from v3.0.0's current code. 
But adding a `try/catch` block around `on-headers` is not that straight forward here. 

**Can encryptSync actually throw in practice?**
The key passed to `encryptSync` is always the output of `getKeyStore(config.secret, true)`, which derives a key using `crypto.hkdfSync` with a fixed output length of 32 bytes. `crypto.createCipheriv('aes-256-gcm', key, iv)` only throws if the key is not exactly 32 bytes, which structurally cannot happen here. Similarly, `crypto.randomBytes(12)` does not throw under any normal operating condition. 

**This is consistent with what v2 had:**
In v2, setCookie inside the on-headers callback was also fully synchronous with no try/catch. jose v2's JWE.encrypt was synchronous, and any error it threw inside the on-headers callback would have surfaced at writeHead time in exactly the same way, not forwarded through the Express error handler. So while this is a behavioural difference from v3.0.0, it is fully consistent with v2, which shipped this way without issues.

**Now why adding try/catch here is non-trivial here:**
The problem is that inside `on-headers`, we're in the middle of writeHead, headers haven't been sent yet, so we can't call `next(e)` meaningfully at that point because the response state is mid-flight. What would we actually do with the caught error? We'd have to store it somewhere and propagate it later, which is complex and arguably worse than just letting the rare uncaught error surface.

In my opinion, we should leave this as is, this matches v2's behaviour exactly.

test/appSession.tests.js
cschetan77: Added - 
1. `should write session cookie even when headers are flushed before res.end() (streaming)`: Streaming test for custom store
2. `should send Set-Cookie even when store.set() fails (cookie written before persist at writeHead time)`: This asserts the behavior, cookie is sent even when store.set() fails. 

The reasoning for 2nd case is same mentioned in https://github.com/auth0/express-openid-connect/pull/838#discussion_r3621142755

lib/appSession.js
cschetan77: yes, this comment is misleading and no longer required. Removed.

nandan-bhat (dismissed):

jd3vi1 (approved):

PR #850, opened 2026-07-09
feat: Add CTE impersonation via Session Transfer Token

cschetan77:
## Summary

Adds **Custom Token Exchange (CTE) — Impersonation via Session Transfer**.

This extends the existing `req.oidc.customTokenExchange` surface with two new methods that implement the initiator role of the session transfer flow: an application requests a short-lived, single-use Session Transfer Token (STT) via CTE, then redirects the user's browser to a target app's login URL carrying that STT, which is redeemed at `/authorize` to establish an ephemeral web session as the subject user with the actor recorded in the `act` claim.

### New methods on `req.oidc`

**`requestSessionTransferToken(options)`**

Performs a token exchange against the `urn:{domain}:session_transfer` audience and returns a `SessionTransferTokenResult`. The SDK handles the protocol plumbing automatically:

- Builds the `audience` from the configured `issuerBaseURL` hostname
- Sources the `actor_token` from the current session `id_token`, checking the id_token's own `exp` claim (with a 30s skew). If expired and a refresh token is available, the session is refreshed automatically; if the refresh returns no new id_token, raises `actor_unavailable`. If expired with no refresh token, raises `actor_unavailable` before any network call. An explicit `actor_token` override is accepted to bypass session sourcing entirely.
- Raises `actor_unavailable` (HTTP 400) client-side before any network call when no actor can be resolved.
- Validates that the AS response carries `issued_token_type === "urn:auth0:params:oauth:token-type:session_transfer_token"` — an unexpected or missing value raises HTTP 500 (`invalid_token_response`).
- The returned STT is **never written to the session** — it is a one-shot handle to be forwarded immediately.

```js
app.post('/impersonate', requiresAuth(), async (req, res) => {
  const result = await req.oidc.requestSessionTransferToken({
    subject_token: req.body.customerToken,
    subject_token_type: 'urn:mycompany:customer-subject',
    extra: { reason: 'Investigating ticket TCK-1234' },
  });
  res.redirect(req.oidc.buildSessionTransferRedirect('https://app.example.com/login', result));
});
```

**`buildSessionTransferRedirect(targetLoginUrl, result, opts?)`**

Pure URL builder — appends `session_transfer_token` (and optionally `organization`) as query parameters to `targetLoginUrl`. Makes no network call and writes nothing to the session. The caller passes the returned string to `res.redirect()`.

> **Security:** `targetLoginUrl` must be a trusted, app-controlled value — never derive it from untrusted input (e.g. a query parameter), as the STT would be forwarded to an attacker-controlled host. The URL must use `https:` — `http:` is accepted only for `localhost`, `127.0.0.1`, and `[::1]` for local development; any other scheme or non-loopback `http:` host throws a `TypeError`.

**Target side (no new SDK methods needed)**

The target app forwards the STT to `/authorize` via `res.oidc.login()`. Two things are required: `authRequired: false` on the middleware (so the middleware doesn't intercept the unauthenticated `/login` request before the custom route runs), and `returnTo: '/'` on the login call (to avoid a redirect loop where the callback redirects back to `/login?session_transfer_token=...`):

```js
app.use(auth({
  // ... your other config
  authRequired: false,
  routes: { login: false },
}));

app.get('/login', async (req, res) => {
  await res.oidc.login({
    authorizationParams: {
      session_transfer_token: req.query.session_transfer_token,
    },
    returnTo: '/',
  });
});
```

### New TypeScript types

- `SessionTransferTokenResult` — result type returned by `requestSessionTransferToken`
- `SessionTransferTokenOptions` — input options for `requestSessionTransferToken`
- `SessionTransferRedirectOptions` — options for `buildSessionTransferRedirect`

### Error codes

| Code | HTTP | Where | When |
|------|------|-------|------|
| `actor_unavailable` | 400 | client-side | No actor passed and no usable session `id_token` (not logged in, expired with no refresh token, refresh returned no new id_token, or no `id_token` in session) |
| `setactor_required` | 400 | server | CTE Action did not call `setActor` |
| `session_transfer_disabled` | 400 | server | Session transfer not enabled on the tenant |
| `invalid_token_response` | 500 | client-side | AS returned an unexpected or missing `issued_token_type` |

### Implementation notes

- Fully additive — no existing types, methods, or behavior changed. All existing `customTokenExchange` tests continue to pass.
- Reuses `validateSubjectToken`, `validateTokenExchangeExtras`, `TOKEN_EXCHANGE_GRANT_TYPE`, and the `genericGrantRequest` CTE machinery.
- `isIdTokenExpired(exp)` added to `lib/utils/sessionExpiry.js`, reusing the existing `SESSION_EXPIRY_LEEWAY` (30s skew). Guards against non-numeric `exp` (fails closed — treats as expired).
- `SESSION_TRANSFER_TOKEN_IDENTIFIER` defined as an internal constant in `lib/context.js` — not exported; the URN is documented in EXAMPLES.md and the type declarations for developers who need to branch on `issued_token_type`.
- `extractOidcError()` added to `lib/utils/oidcError.js` — extracts `error_description` from RFC 6749 AS error responses and the `cause` chain from `openid-client` v6 `ClientError` objects. Used in `requestSessionTransferToken` so consumers see the actual AS error (e.g. `"setActor is required..."`) rather than the generic `openid-client` wrapper message.
- `examples/cte-session-transfer.js` — runnable two-server example demonstrating the full initiator → target flow.

## Test plan

- [x] 35 tests in `test/sessionTransferToken.tests.js` covering: actor resolution (session sourcing, explicit override, blank actor, no session, no id_token in session, expired id_token with and without refresh token, refresh returning no new id_token, non-numeric exp), `subject_token` validation, audience construction, optional params (`scope`, `organization`, `extra`), result shape (STT in `session_transfer_token`, no `access_token`, no `act`), `issued_token_type` validation at exchange (unexpected and missing both throw 500), no-persistence guarantee, server error code mapping, and `buildSessionTransferRedirect` URL construction (including https enforcement and all three loopback variants).
- [x] All existing tests pass without modification.
- [x] End-to-end smoke test passed: STT minted, forwarded to target app, redeemed at `/authorize`, impersonation session established with correct `sub` (customer) and `act.sub` (agent). Auth0 logs confirmed `triggered_by_session_transfer_token: true`.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -12,8 +12,9 @@
 10. [Use a custom session store](#10-use-a-custom-session-store)
 11. [Back-Channel Logout](#11-back-channel-logout)
 12. [Custom Token Exchange](#12-custom-token-exchange)
-13. [Use a proxy for OIDC requests](#13-use-a-proxy-for-oidc-requests)
-14. [Session expiry from upstream IdP (IPSIE `session_expiry`)](#14-session-expiry-from-upstream-idp-ipsie-session_expiry)
+13. [Impersonation via Session Transfer Token](#13-impersonation-via-session-transfer-token)
+14. [Use a proxy for OIDC requests](#14-use-a-proxy-for-oidc-requests)
+15. [Session expiry from upstream IdP (IPSIE `session_expiry`)](#15-session-expiry-from-upstream-idp-ipsie-session_expiry)
 
 ## 1. Basic setup
 
@@ -450,7 +451,117 @@ const tokenSet = await req.oidc.customTokenExchange({
 });
 ```
 
-## 13. Use a proxy for OIDC requests
+## 13. Impersonation via Session Transfer Token
+
+Custom Token Exchange Impersonation via Session Transfer lets a support or admin application log a user into a target web application as a customer — so a support engineer can reproduce the customer's exact experience without knowing their password. The agent is recorded in the `act` claim on the impersonated session, making every impersonation auditable.
+
+The flow involves two roles:
+
+- **Initiator** — your support/admin app. It requests a short-lived, single-use Session Transfer Token (STT) and redirects the user's browser to the target app carrying the STT.
+- **Target** — the customer's web app. It forwards the STT to `/authorize`, where Auth0 redeems it and establishes a session as the customer.
+
+> **Prerequisites (configured via Management API / Dashboard):**
+>
+> - Feature flag `cte_session_transfer_token` must be enabled on the tenant.
+> - The issuing (initiator) client needs `can_create_session_transfer_token: true` and a Custom Token Exchange Action that calls `setActor`.
+> - The redeeming (target) client needs `allow_delegated_access: true` and `"query"` in its allowed authentication methods.
+
+### Initiator: requesting an STT and redirecting
+
+The user must be logged in to the initiator app — the SDK sources the actor from the session's ID token by default (refreshing it automatically if expired). Run your own authorization check before calling the SDK.
+
+```js
+const { auth, requiresAuth } = require('express-openid-connect');
+
+app.post('/impersonate', requiresAuth(), async (req, res, next) => {
+  try {
+    // Your own proof of which customer to impersonate — validated by your Action.
+    const { customerToken, customerTokenType } = req.body;
+
+    const result = await req.oidc.requestSessionTransferToken({
+      subject_token: customerToken,
+      subject_token_type: customerTokenType,
+      // Optional: pass custom context to your Action via event.request.body
+      extra: { reason: 'Investigating ticket TCK-1234' },
+    });
+
+    // targetLoginUrl must be a trusted, app-controlled value — never derived from
+    // untrusted input such as a returnTo param, or the STT could leak to an attacker host.
+    const redirectUrl = req.oidc.buildSessionTransferRedirect(
+      'https://app.example.com/auth/login',
+      result,
+    );
+
+    res.redirect(redirectUrl);
+  } catch (err) {
+    // err.error === 'actor_unavailable'        — user not logged in or session expired
+    // err.error === 'setactor_required'        — Action did not call setActor
+    // err.error === 'session_transfer_disabled' — tenant feature flag is off
+    next(err);
+  }
+});
+```
+
+The STT is opaque, single-use, and lives ~60 seconds. The SDK never stores it — hand it straight to `buildSessionTransferRedirect` and discard it.
+
+If the customer belongs to an organization, forward it on the redirect:
+
+```js
+const redirectUrl = req.oidc.buildSessionTransferRedirect(
+  'https://app.example.com/auth/login',
+  result,
+  { organization: 'org_globex' },
+);
+```
+
+To supply the acting party explicitly instead of using the session ID token:
+
+```js
+const result = await req.oidc.requestSessionTransferToken({
+  subject_token: customerToken,
+  subject_token_type: customerTokenType,
+  actor_token: agentIdToken,
+  actor_token_type: 'urn:ietf:params:oauth:token-type:id_token',
+});
+```
+
+### Target: redeeming the STT
+
+On the target app, forward the `session_transfer_token` query parameter (and `organization` when present) to `/authorize` through the existing `res.oidc.login()` call — no new SDK methods required:
+
+```js
+app.get('/auth/login', async (req, res) => {
+  const authorizationParams = {};
+
+  if (req.query.session_transfer_token) {
+    authorizationParams.session_transfer_token =
+      req.query.session_transfer_token;
+  }
+  if (req.query.organization) {
+    authorizationParams.organization = req.query.organization;
+  }
+
+  await res.oidc.login({ authorizationParams });
+});
+```
+
+The established session is short-lived (hard-capped at 2 hours) and cannot mint a refresh token.
+
+### Reading the `act` claim
+
+Once the impersonation session is established, the acting party is available as `req.oidc.user.act`:
+
+```js
+app.get('/dashboard', requiresAuth(), (req, res) => {
+  const actor = req.oidc.user.act; // { sub: 'support-agent-007' } when impersonated
+  if (actor) {
+    // Render an impersonation banner so the agent knows they are acting as the customer.
+  }
+  res.render('dashboard');
+});
+```
+
+## 14. Use a proxy for OIDC requests
 
 If you need to route all OIDC HTTP requests (discovery, token, userinfo, etc.) through a proxy, use the `customFetch` option with `undici`'s `ProxyAgent`:
 
@@ -473,7 +584,7 @@ app.use(
 
 The SDK wraps your `customFetch` function to add required headers (User-Agent, Auth0-Client telemetry) before making requests.
 
-## 14. Session expiry from upstream IdP (IPSIE `session_expiry`)
+## 15. Session expiry from upstream IdP (IPSIE `session_expiry`)
 
 When an upstream IdP supports the IPSIE SL1 spec, it can include a `session_expiry` claim in the ID token — an absolute Unix timestamp (seconds) marking the latest moment the IdP considers the session valid.
 
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -453,18 +453,16 @@ const tokenSet = await req.oidc.customTokenExchange({
 
 ## 13. Impersonation via Session Transfer Token
 
-Custom Token Exchange Impersonation via Session Transfer lets a support or admin application log a user into a target web application as a customer — so a support engineer can reproduce the customer's exact experience without knowing their password. The agent is recorded in the `act` claim on the impersonated session, making every impersonation auditable.
+Custom Token Exchange Impersonation via Session Transfer lets a support or admin application log a user into a target web application as a customer — so a support engineer can reproduce the customer's exact experience without knowing their password. The initiator is recorded in the `act` claim on the impersonated session, making every impersonation auditable.
 
 The flow involves two roles:
 
 - **Initiator** — your support/admin app. It requests a short-lived, single-use Session Transfer Token (STT) and redirects the user's browser to the target app carrying the STT.
-- **Target** — the customer's web app. It forwards the STT to `/authorize`, where Auth0 redeems it and establishes a session as the customer.
+- **Target** — the customer's web app. It forwards the STT to `/authorize`, where Auth0 redeems it and establishes an ephemeral session as the customer, recording the initiator in the `act` (actor) claim.
 
-> **Prerequisites (configured via Management API / Dashboard):**
->
-> - Feature flag `cte_session_transfer_token` must be enabled on the tenant.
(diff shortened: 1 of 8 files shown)


index.d.ts
```diff
@@ -127,6 +127,10 @@ interface ActClaim {
   [key: string]: unknown;
 }
 
+/** @internal Token-type URI constants for use with STT and CTE methods. */
+export const SESSION_TRANSFER_TOKEN_IDENTIFIER: 'urn:auth0:params:oauth:token-type:session_transfer_token';
```
kishore7snehil: These two are declared as `export const`, so a TypeScript user can write `import { SESSION_TRANSFER_TOKEN_IDENTIFIER } from 'express-openid-connect'` and it compiles, but the package's runtime entry only exports `auth`, `requiresAuth`, `attemptSilentLogin`, and `SessionExpiredError`, so at runtime they come back `undefined`. They're also tagged `@internal`, which reads as "not meant for consumers" and pulls the other way.

Shall we either export them for real from the main entry, or drop them from the types if they're internal only?

lib/context.js
```diff
@@ -442,6 +446,168 @@ class RequestContext {
...
+        } catch {
+          throw createError(
+            400,
+            'No actor available: session expired and could not be refreshed.',
+            { error: 'actor_unavailable' },
+          );
+        }
+      }
+
+      resolvedActorToken = this.idToken;
```
kishore7snehil, line 530: After the refresh runs we read `this.idToken` and send it as the actor, but `refresh` sets `id_token` to `newTokenSet.id_token || oldTokenSet.id_token`, so when the refresh response carries no new id_token the old expired one is kept and forwarded anyway. The server then rejects it on the `exp` check, which is the exact case the refresh here is meant to prevent.

Can we re-check expiry on the resolved id_token after the refresh and fail with `actor_unavailable` if it's still expired? There's also no test for the refresh-returns-no-id_token path, so a quick one would lock it in.

lib/context.js
```diff
@@ -442,6 +446,168 @@ class RequestContext {
...
+    try {
+      const exchanged = await oidcClient.genericGrantRequest(
+        configuration,
+        TOKEN_EXCHANGE_GRANT_TYPE,
+        parameters,
+      );
+
+      return {
+        session_transfer_token: exchanged.access_token,
+        issued_token_type: exchanged.issued_token_type || '',
```
kishore7snehil: This defaults `issued_token_type` to `''` when the response omits it, but `buildSessionTransferRedirect` only accepts a result whose `issued_token_type` is exactly the session-transfer URN. So a result this method happily returns can't be handed to the redirect helper, and the empty-string case is even pinned by a test. It also means this method never confirms the response is actually an STT, so a non-STT `access_token` would come back in `session_transfer_token` for anyone doing their own redirect.

Shall we check `issued_token_type` here at the source and fail the exchange if it isn't the STT URN, rather than catching it later at redirect time?

lib/context.js
```diff
@@ -442,6 +446,168 @@ class RequestContext {
...
+      !result.session_transfer_token
+    ) {
+      throw new TypeError(
+        'result must be a SessionTransferTokenResult with a session_transfer_token',
+      );
+    }
+
+    let redirectUrl;
+    try {
+      redirectUrl = new URL(targetLoginUrl);
```
kishore7snehil, line 612: `new URL(targetLoginUrl)` accepts any scheme, so an `http:` (or `javascript:`/`data:`) target parses fine and the STT gets appended as a query param. The JSDoc already says to use `https:` in production, and since this helper is the one place that hardening can live, an `http:`/`https:` allowlist (with a `localhost` exception for dev) would make that guidance actually hold.

lib/utils/sessionExpiry.js
```diff
@@ -50,9 +50,19 @@ function isSessionExpiryInPast(sessionExpiresAt, issuedAt) {
   return isSessionExpiryReached(sessionExpiresAt, reference);
 }
 
+/**
+ * Returns whether an ID token's `exp` claim is expired, applying a forward skew so a token
+ * about to expire is treated as already expired and refreshed before being sent as an actor token.
+ * @param {number} exp - the `exp` claim from the decoded ID token
+ */
+function isIdTokenExpired(exp) {
+  return exp <= epoch() + SESSION_EXPIRY_LEEWAY;
```
kishore7snehil, line 60: This compares `exp` without checking it's a number, so a missing or malformed `exp` makes `undefined <= number` evaluate false and the token reads as not expired, skipping the refresh and getting forwarded as the actor. The other helpers in this file already guard their inputs for plausibility, so this one is the odd one out.

Can we fail closed here, treating a non-numeric `exp` as expired so an unknown expiry forces a refresh or `actor_unavailable` rather than sending a token we can't vouch for?

lib/context.js
```diff
@@ -442,6 +446,168 @@ class RequestContext {
...
+      const idToken = this.idToken;
+      if (!idToken) {
+        throw createError(
+          400,
+          'No actor available: session exists but contains no id_token. Ensure the session was established with the openid scope, or pass an explicit actor_token.',
+          { error: 'actor_unavailable' },
+        );
+      }
+
+      const { exp } = ts.claims();
```
kishore7snehil: `ts.claims()` can come back undefined for an id_token that won't decode, and this destructure would then throw a `TypeError` that surfaces as a 500, stepping outside the `actor_unavailable` 400 contract this method documents. The checks just above guard `!ts` and `!idToken` but not this case.

Shall we read it as `ts.claims() || {}` so an undecodable token stays within the documented error path?

EXAMPLES.md
```diff
@@ -450,7 +451,117 @@ const tokenSet = await req.oidc.customTokenExchange({
...
+  } catch (err) {
+    // err.error === 'actor_unavailable'        — user not logged in or session expired
+    // err.error === 'setactor_required'        — Action did not call setActor
+    // err.error === 'session_transfer_disabled' — tenant feature flag is off
+    next(err);
+  }
+});
+```
+
+The STT is opaque, single-use, and lives ~60 seconds. The SDK never stores it — hand it straight to `buildSessionTransferRedirect` and discard it.
```
kishore7snehil: The guide covers the one-shot nature, the trusted-URL warning, and reading `act` well, but two things a developer will hit aren't called out. There's no note that an explicit `actor_token` must be an unexpired, asymmetrically-signed JWT (an HS256 or expired token is rejected server-side), and nothing tells them to branch on `issued_token_type` even though the redirect helper hard-requires that exact URN.

Can we add a short note on both, so the actor-token constraint and the field-to-branch-on are visible without reading the source?

kishore7snehil (changes requested):

index.d.ts
cschetan77: Considered making `SESSION_TRANSFER_TOKEN_IDENTIFIER` a real public export, but decided against it.
Removed both from the type declarations.

lib/context.js
cschetan77: yes, that's valid point. 
Fixed, after `refresh.call(this)` we now re-read `tokenSet.claims().exp` and throw `actor_unavailable` if the id_token is still expired. 
Added a test that covers the path: refresh grant succeeds (200) but returns no id_token, causing the old expired token to be kept, the method now fails with actor_unavailable rather than forwarding it.

lib/context.js
cschetan77: yes, that makes sense and a better design. 
Done. The `issued_token_type` is now validated immediately after the exchange returns. A missing or unexpected value throws a 500, the request itself was valid, but the upstream returned an unexpected response shape. 
2 unit tests have been added to assert the same.

lib/context.js
cschetan77: Done. `buildSessionTransferRedirect` now rejects any targetLoginUrl that isn't `https:`, with an exception for `http:` on `localhost`, `127.0.0.1`, and `[::1]` to support local development. Any other scheme (http: on a non-loopback host, ftp:, javascript:, data:, etc.) throws a TypeError. Aligns with the same check in auth0-server-js. 
Five tests added covering the rejection cases and all three loopback variants.

lib/utils/sessionExpiry.js
cschetan77: yes, we should certainly guard against missing or malformed exp. 
Done. Added a `typeof exp !== 'number' || !Number.isFinite(exp)` guard, a missing, NaN, or non-numeric exp now returns true (expired) rather than falling through to the <= comparison where undefined <= number evaluates false. 
A test covers the path: a session id_token with exp: 'not-a-number' is treated as expired and fails with actor_unavailable when no refresh token is available.

Also, considered reusing `isPlausibleUnixSeconds` (already in this file) but kept them separate, isPlausibleUnixSeconds was purpose-built for the session_expiry claim and has stricter constraints (integer-only, upper bound) that would bleed the wrong semantics into id_token expiry checking.

lib/context.js
cschetan77: Done.

EXAMPLES.md
cschetan77: Added both notes:

1. A callout after the explicit actor_token example, the token must be unexpired and RS256/PS256 signed, HS256 and expired tokens are rejected server-side. Auth0 session ID tokens already satisfy both, but the constraint matters if the token comes from elsewhere.
2. A callout after the main result handling, branch on `result.issued_token_type`, not `result.token_type`. token_type is "N_A" for STT responses, buildSessionTransferRedirect requires the exact `issued_token_type URN`.

EXAMPLES.md
```diff
@@ -450,7 +451,121 @@ const tokenSet = await req.oidc.customTokenExchange({
...
+Custom Token Exchange Impersonation via Session Transfer lets a support or admin application log a user into a target web application as a customer — so a support engineer can reproduce the customer's exact experience without knowing their password. The agent is recorded in the `act` claim on the impersonated session, making every impersonation auditable.
+
+The flow involves two roles:
+
+- **Initiator** — your support/admin app. It requests a short-lived, single-use Session Transfer Token (STT) and redirects the user's browser to the target app carrying the STT.
+- **Target** — the customer's web app. It forwards the STT to `/authorize`, where Auth0 redeems it and establishes a session as the customer.
+
+> **Prerequisites (configured via Management API / Dashboard):**
+>
+> - Feature flag `cte_session_transfer_token` must be enabled on the tenant.
```
kishore7snehil: We can avoid calling out the flag name as it's an internal detail

kishore7snehil (changes requested):

EXAMPLES.md
cschetan77: Agree. Removed all the occurrences of internal details.

kishore7snehil (dismissed):
LGTM!

examples/cte-session-transfer.js
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    // returnTo: '/' prevents an infinite redirect loop: without it the SDK
+    // stores /login?session_transfer_token=... as the post-callback destination,
+    // which triggers another login cycle after the callback completes.
+    await res.oidc.login({ authorizationParams, returnTo: '/' });
+  } catch (err) {
+    next(err);
+  }
+});
```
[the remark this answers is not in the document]
cschetan77: Internal test file.

amitsingh05667 (approved):

PR #863, opened 2026-07-24
feat(jar): JWT-Secured Authorization Requests (JAR) support

jd3vi1:
## Summary

Adds support for [JWT-Secured Authorization Requests (JAR, RFC 9101)](https://www.rfc-editor.org/rfc/rfc9101.html). When `requestObjectSigningKey` is configured, the SDK signs all authorization parameters into a JWT and sends it as the `request` parameter, so the `/authorize` redirect carries only `client_id` and `request`.

## What's included

- **`requestObjectSigningKey`** — private key used to sign the request object. Accepts the same formats as `clientAssertionSigningKey` (PEM string, Buffer, KeyObject, JWK, CryptoKey).
- **`requestObjectSigningAlg`** — required whenever `requestObjectSigningKey` is set. Web Crypto algorithm names are not valid JWA `alg` values, so an explicit JWA algorithm is required (a shared `ASYMMETRIC_SIGNING_ALGS` list is used, matching `clientAssertionSigningAlg`).
- **`requestObjectSigningKeyId`** — optional `kid` header on the request object JWT.
- Works transparently on top of the existing PAR support. When PAR is enabled, the signed request object is POSTed to `/oauth/par`, which is the recommended FAPI pattern.
- The request object's `aud` is set to the issuer identifier advertised in the discovery document (which may differ from `issuerBaseURL`, e.g. a trailing slash), as JAR requires.

## Usage

```js
app.use(
  auth({
    authorizationParams: { response_type: 'code' },
    requestObjectSigningKey: fs.readFileSync('./request-object-key.pem'),
    requestObjectSigningAlg: 'RS256',
    pushedAuthorizationRequests: true, // recommended
  }),
);
```

## Tests

- Unit tests for `buildRequestObject` (header, `kid`, standard claims, `aud`), config validation (alg required with key), and the login redirect (`request` param present, PAR combination).
- An end-to-end test covering the combined PAR + JAR flow.
- Type tests in `index.test-d.ts`.
- Documentation in `EXAMPLES.md` and a runnable example at `examples/jar.js`.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -15,6 +15,7 @@
 13. [Impersonation via Session Transfer Token](#13-impersonation-via-session-transfer-token)
 14. [Use a proxy for OIDC requests](#14-use-a-proxy-for-oidc-requests)
 15. [Session expiry from upstream IdP (IPSIE `session_expiry`)](#15-session-expiry-from-upstream-idp-ipsie-session_expiry)
+16. [JWT-Secured Authorization Requests (JAR)](#16-jwt-secured-authorization-requests-jar)
 
 ## 1. Basic setup
 
@@ -667,3 +668,30 @@ app.get('/status', (req, res) => {
 ### Upgrading existing apps
 
 Once your IdP starts emitting `session_expiry`, `req.appSession` can be `null` for a previously logged-in user once the ceiling is reached. If your code assumed the session always exists after login, add a null check. Sessions created before the upgrade (or through connections without the claim) have no `sessionExpiresAt` and behave exactly as before.
+
+## 16. JWT-Secured Authorization Requests (JAR)
+
+[JAR](https://www.rfc-editor.org/rfc/rfc9101.html) (part of Auth0's Highly Regulated Identity feature set) signs all authorization parameters into a JWT and sends them as the `request` parameter, so the `/authorize` redirect carries only `client_id` and `request`. Set `requestObjectSigningKey` to enable it. `requestObjectSigningAlg` is always required, since Web Crypto algorithm names are not valid JWA `alg` values.
+
+```js
+app.use(
+  auth({
+    authorizationParams: {
+      response_type: 'code',
+    },
+    // Sign the authorization request object.
+    requestObjectSigningKey: fs.readFileSync('./request-object-key.pem'),
+    requestObjectSigningAlg: 'RS256',
+    // Optional: set a `kid` in the request object's JWT header.
+    // requestObjectSigningKeyId: 'my-key-id',
+
+    // Recommended for FAPI: combine with PAR so the signed request object is
+    // POSTed to /oauth/par and only an opaque request_uri is exposed in the URL.
+    pushedAuthorizationRequests: true,
+  }),
+);
+```
+
+The request object's `aud` is set to the issuer identifier advertised in the discovery document (which may differ from `issuerBaseURL`, e.g. a trailing slash), as JAR requires. `requestObjectSigningKey` accepts the same key formats as `clientAssertionSigningKey` (PEM string, Buffer, KeyObject, JWK, CryptoKey).
+
+Full example at [jar.js](./examples/jar.js), to run it: `npm run start:example -- jar`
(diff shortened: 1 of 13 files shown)


jd3vi1:
Addressed the nits in 7ff0457:

1. **`aud` fallback removed.** `buildRequestObject` now requires the discovered issuer as its `audience` argument and throws if it is missing, so the `issuerBaseURL` fallback (the mismatch that triggers `invalid_request_object`) is gone. The unit tests now pass the discovered issuer explicitly and assert `aud` equals it.
2. **Non-PAR redirect coverage added.** New test asserts the non-PAR authorize redirect carries only `client_id` and `request`.
3. **Empty-string key.** `requestObjectSigningKey` now rejects `''` at config validation, so the "is set" check is consistent between the schema and the login-side guard (previously `''` forced `alg` required but silently skipped JAR).

Full suite and tsd green.

examples/run_example.js
```diff
@@ -12,10 +12,15 @@ if (!process.env.CLIENT_ID) {
   console.log(
     'Starting a mock authorization server. You can login with any credentials.',
   );
+  // The hri and jar examples use a registered client that supports
+  // private_key_jwt + JAR + PAR. All other examples use the default client.
+  const clientId = ['hri.js', 'jar.js'].includes(example)
```
gyaneshgouraw-okta: Hey @jd3vi1 i don't see any file `hri.js` in codebase. Can you please check ?

index.d.ts
```diff
@@ -871,7 +870,12 @@ interface ConfigParams {
   /**
    * String value for the client's authentication method. Default is `none` when using response_type='id_token', `private_key_jwt` when using a `clientAssertionSigningKey`, otherwise `client_secret_basic`.
    */
-  clientAuthMethod?: string;
+  clientAuthMethod?:
+    | 'client_secret_basic'
```
gyaneshgouraw-okta: @jd3vi1 this may result into compile time error for existing users, also i believe it's not directly related to JAR changes.

jd3vi1: agreed, reverting

EXAMPLES.md
```diff
@@ -667,3 +668,30 @@ app.get('/status', (req, res) => {
...
+    // Recommended for FAPI: combine with PAR so the signed request object is
+    // POSTed to /oauth/par and only an opaque request_uri is exposed in the URL.
+    pushedAuthorizationRequests: true,
+  }),
+);
+```
+
+The request object's `aud` is set to the issuer identifier advertised in the discovery document (which may differ from `issuerBaseURL`, e.g. a trailing slash), as JAR requires. `requestObjectSigningKey` accepts the same key formats as `clientAssertionSigningKey` (PEM string, Buffer, KeyObject, JWK, CryptoKey).
+
+Full example at [jar.js](./examples/jar.js), to run it: `npm run start:example -- jar`
```
gyaneshgouraw-okta, line 715: Hey @jd3vi1 based on the example added for this change, the accepted command to run the example app would be 
```js
npm run start:example -- jar.js
or 
npm run start:example -- jar
```

Can you please check ?

EXAMPLES.md
```diff
@@ -667,3 +668,30 @@ app.get('/status', (req, res) => {
 ### Upgrading existing apps
 
 Once your IdP starts emitting `session_expiry`, `req.appSession` can be `null` for a previously logged-in user once the ceiling is reached. If your code assumed the session always exists after login, add a null check. Sessions created before the upgrade (or through connections without the claim) have no `sessionExpiresAt` and behave exactly as before.
+
+## 16. JWT-Secured Authorization Requests (JAR)
+
+[JAR](https://www.rfc-editor.org/rfc/rfc9101.html) (part of Auth0's Highly Regulated Identity feature set) signs all authorization parameters into a JWT and sends them as the `request` parameter, so the `/authorize` redirect carries only `client_id` and `request`. Set `requestObjectSigningKey` to enable it. `requestObjectSigningAlg` is always required, since Web Crypto algorithm names are not valid JWA `alg` values.
```
gyaneshgouraw-okta, line 674: @jd3vi1 we should showcase or provide some reference how users can get/generate `requestObjectSigningKey` and also reference for allowed `requestObjectSigningAlg`

EXAMPLES.md
jd3vi1: i've made it agnostic of the extension, both should work now.

EXAMPLES.md
jd3vi1: added

gyaneshgouraw (approved):

PR #865, opened 2026-07-24
feat(mtls): mTLS (RFC 8705) client authentication

jd3vi1:
## Summary

Adds mTLS ([RFC 8705](https://www.rfc-editor.org/rfc/rfc8705)) client authentication. When enabled, the SDK authenticates to the token endpoint with a TLS client certificate instead of a client secret, and issued access tokens carry a `cnf.x5t#S256` claim binding them to the certificate (certificate-bound tokens).

Independent of #863 (JAR); both branch from `master`.

## What's included

- **`useMtls: boolean`** (or `AUTH0_MTLS=true`) — single opt-in. Internally uses `TlsClientAuth()` and sets `use_mtls_endpoint_aliases: true` on the client metadata, so token, refresh, revocation, userinfo, and PAR requests route to the server's `mtls_endpoint_aliases`. There is no separate CA-signed vs self-signed option; that distinction is an authorization-server concern.
- **`customFetch` is required** with `useMtls`, and must present the client certificate at the TLS layer (e.g. an `undici` `Agent` with `connect: { key, cert }`). The certificate is never configured through the SDK directly.
- **New `MtlsError` / `MtlsErrorCode`** (exported), thrown at construction / discovery time:
  - `mtls_requires_custom_fetch` — `useMtls` without a `customFetch`.
  - `mtls_incompatible_client_auth` — `useMtls` combined with `clientSecret`, `clientAssertionSigningKey`, or an explicit `clientAuthMethod`.
  - `mtls_endpoint_aliases_missing` — the discovery document does not advertise `mtls_endpoint_aliases.token_endpoint` (fails fast rather than a silent runtime `invalid_client`).
- A startup warning when `useMtls` is used against a canonical `*.auth0.com` issuer, since mTLS requires a custom domain.

## Usage

```js
const { Agent, fetch: undiciFetch } = require('undici');

const tlsAgent = new Agent({
  connect: {
    cert: fs.readFileSync('./client.crt'),
    key: fs.readFileSync('./client.key'),
  },
});

app.use(
  auth({
    issuerBaseURL: 'https://auth.your-domain.com', // custom domain
    authorizationParams: { response_type: 'code', audience: 'https://your-api/' },
    useMtls: true,
    customFetch: (url, options) => undiciFetch(url, { ...options, dispatcher: tlsAgent }),
  }),
);
```

## Tests

- Unit tests for `use_mtls_endpoint_aliases` metadata, aliases surfaced in server metadata, the alias-missing `MtlsError`, token grants routing to the mTLS alias, and all construction guards (`customFetch` required, incompatible client auth including an explicit `clientAuthMethod`, `AUTH0_MTLS` env, rejection of the `tls_client_auth` string).
- Type tests for the `useMtls` / `customFetch` surface and the `MtlsError` / `MtlsErrorCode` shape.
- Validated end to end against a live tenant with a self-managed-certs custom domain and a local edge proxy: login succeeds and the access token carries `cnf.x5t#S256` matching the client certificate thumbprint, including across refresh.
- Documentation in `EXAMPLES.md` and an example at `examples/mtls.js`.

### DIFF
--- a/examples/mtls.js
+++ b/examples/mtls.js
@@ -2,14 +2,17 @@ const express = require('express');
 const { auth } = require('../');
 const { Agent, fetch: undiciFetch } = require('undici');
 const fs = require('fs');
+const path = require('path');
 
 const app = express();
 
 // mTLS (Mutual TLS, RFC 8705) client authentication demo.
 //
 // With `useMtls: true` the SDK authenticates to the token endpoint with a TLS
-// client certificate instead of a client secret, and routes token/refresh/
-// revocation/userinfo/PAR requests to the server's `mtls_endpoint_aliases`.
+// client certificate instead of a client secret. Requests are sent to the
+// server's `mtls_endpoint_aliases` for each endpoint the server advertises an
+// alias for; an endpoint without an alias is sent over the standard (non-mTLS)
+// channel, so the tenant must advertise aliases for every endpoint you use.
 // Issued access tokens carry a `cnf.x5t#S256` claim binding them to the
 // certificate (certificate-bound tokens).
 //
@@ -26,10 +29,26 @@ const app = express();
 //
 // `AUTH0_MTLS=true` can be used instead of `useMtls: true`.
 
+// Resolve the certificate and key relative to this file so the example works
+// regardless of the caller's working directory. Fail with a message naming the
+// expected files rather than an opaque ENOENT.
+const readCertFile = (name) => {
+  const file = path.join(__dirname, name);
+  try {
+    return fs.readFileSync(file);
+  } catch (e) {
+    throw new Error(
+      `mTLS example: could not read "${file}". Place your client certificate ` +
+        `("client.crt") and private key ("client.key") next to this example. ` +
+        `(${e.code || e.message})`,
+    );
+  }
+};
+
 const tlsAgent = new Agent({
   connect: {
-    cert: fs.readFileSync('./client.crt'),
-    key: fs.readFileSync('./client.key'),
+    cert: readCertFile('client.crt'),
+    key: readCertFile('client.key'),
   },
 });
 
--- a/examples/mtls.js
+++ b/examples/mtls.js
@@ -1,0 +1,61 @@
+const express = require('express');
+const { auth } = require('../');
+const { Agent, fetch: undiciFetch } = require('undici');
+const fs = require('fs');
+
+const app = express();
+
+// mTLS (Mutual TLS, RFC 8705) client authentication demo.
+//
+// With `useMtls: true` the SDK authenticates to the token endpoint with a TLS
+// client certificate instead of a client secret, and routes token/refresh/
+// revocation/userinfo/PAR requests to the server's `mtls_endpoint_aliases`.
+// Issued access tokens carry a `cnf.x5t#S256` claim binding them to the
+// certificate (certificate-bound tokens).
+//
+// The certificate is presented at the TLS layer by your customFetch, never by
+// the SDK. Node's global fetch ignores the `agent` option, so the cert must ride
+// on an undici Agent's `connect` options via the `dispatcher`.
+//
+// Prerequisites (see the mTLS docs):
+//   - A custom domain with self-managed certs (does NOT work on *.auth0.com).
+//   - mTLS endpoint aliases enabled on the tenant.
+//   - App Credentials > Authentication Method set to mTLS, with the client cert
+//     uploaded.
+//   - No clientSecret / clientAssertionSigningKey (mutually exclusive with mTLS).
+//
+// `AUTH0_MTLS=true` can be used instead of `useMtls: true`.
+
+const tlsAgent = new Agent({
+  connect: {
+    cert: fs.readFileSync('./client.crt'),
+    key: fs.readFileSync('./client.key'),
+  },
+});
+
+app.use(
+  auth({
+    // Point issuerBaseURL at your custom domain, not the *.auth0.com host.
+    issuerBaseURL: 'https://auth.your-domain.com',
+    authRequired: false,
+    authorizationParams: {
+      response_type: 'code',
+      audience: 'https://your-api/',
+      scope: 'openid profile email offline_access',
+    },
+
+    useMtls: true,
+    customFetch: (url, options) =>
+      undiciFetch(url, { ...options, dispatcher: tlsAgent }),
+  }),
+);
+
+app.get('/', (req, res) => {
+  if (req.oidc.isAuthenticated()) {
+    res.send(`hello ${req.oidc.user.sub} <a href="/logout">logout</a>`);
+  } else {
+    res.send('<a href="/login">login</a>');
+  }
+});
+
+module.exports = app;
--- a/lib/config.js
+++ b/lib/config.js
@@ -470,6 +470,19 @@ module.exports.get = function (config = {}) {
           'It will not work with canonical *.auth0.com domains.',
       );
     }
+
+    // mTLS only takes effect on the token endpoint, which is reached only when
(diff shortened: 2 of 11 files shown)


jd3vi1:
Both majors fixed in 4fc699c:

1. **`tls_client_auth` removed from the Joi `.valid()` list.** Confirmed your finding that a `.default()` return is not validated against `.valid()`, so the `useMtls` resolver still returns `tls_client_auth` and mTLS keeps working, while a JS caller passing the string directly (bypassing every `useMtls` guard) is now rejected.
2. **Explicit clientAuthMethod + useMtls now rejected.** Added a guard that throws `MTLS_INCOMPATIBLE_CLIENT_AUTH` when `useMtls` is combined with an explicit non-mTLS `clientAuthMethod` (captured from the raw config before Joi's default resolver fills it), closing the `use_mtls_endpoint_aliases` + `ClientSecretBasic(undefined)` contradiction.
3. **Test fixes.** Replaced the mislabeled test (it passed `self_signed_tls_client_auth`, which was never in the list) with an explicit `tls_client_auth` rejection test, kept a `self_signed_tls_client_auth` rejection test, and added both incompatible-method cases (`client_secret_basic` and `none`).

On the `cnf.x5t#S256` debug notice: it is `debug()`-only output with no assertable side effect and the suite has no debug-capture harness, so I left it untested rather than add brittle plumbing for a minor.

Verified: both fixes exercised end to end; full suite (402), tsd, e2e, and lint all green.

examples/mtls.js
```diff
@@ -0,0 +1,61 @@
...
+  }),
+);
+
+app.get('/', (req, res) => {
+  if (req.oidc.isAuthenticated()) {
+    res.send(`hello ${req.oidc.user.sub} <a href="/logout">logout</a>`);
+  } else {
+    res.send('<a href="/login">login</a>');
+  }
+});
```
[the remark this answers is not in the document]
jd3vi1: it's an example

jd3vi1:
Restructured: this PR is now independent of #863 (JAR) and branches directly from `master`. The JWE PR (#864) was closed (JWE access-token decryption is a Resource Server concern, tracked separately), so mTLS no longer stacks on it.

Heads up on merge order: #863 (JAR) and this PR both touch a few of the same files, so whichever merges second will need a small conflict resolution in `EXAMPLES.md`, `index.test-d.ts`, `lib/config.js`, `test/client.tests.js`, and `test/config.tests.js`. All conflicts are "both PRs appended their own block at the same spot" (a config guard, a doc section, test `describe` blocks) with no overlapping logic, so the resolution is keep-both. Nothing in JAR and mTLS contradicts each other.

lib/context.js
```diff
@@ -836,6 +836,26 @@ class ResponseContext {
           : undefined,
       };
 
+      // When mTLS client authentication is in use, a certificate-bound access
+      // token carries a `cnf.x5t#S256` claim (RFC 8705 §3). Its absence on a JWT
+      // access token means sender-constraining is not active (e.g. the resource
+      // server is not configured for mTLS token binding). Warn so the developer
+      // is not surprised by unbound tokens. Opaque/encrypted tokens are skipped.
+      if (config.useMtls && session.access_token) {
```
Piyush-85: The `cnf.x5t#S256` binding check runs only here in the callback path. The `refresh()` function (line 200) assigns a new `access_token` into the session (line 230) but never re-checks it, so a token that loses its certificate binding on refresh produces no warning. The Testing Doc's "Token Refresh" section verifies `cnf.x5t#S256` on the refreshed token.

Suggest extracting this into a small helper and calling it from `refresh()` too.

lib/context.js
```diff
@@ -836,6 +836,26 @@ class ResponseContext {
...
+          const decoded = decodeJwt(session.access_token);
+          if (!decoded.cnf?.['x5t#S256']) {
+            debug(
+              'mTLS is enabled but the access token has no "cnf.x5t#S256" claim; ' +
+                'it is not certificate-bound. Enable Token Sender-Constraining ' +
+                '(mTLS) on the resource server to bind tokens to the certificate.',
+            );
+          }
+        } catch {
+          // Opaque (non-JWT) access token — cannot inspect for cnf, skip.
```
Piyush-85: Minor: JWE access tokens (issued when no audience is set) also throw in `decodeJwt` and get skipped here, but the Testing Doc expects a warning for the JWE case. Flagging for awareness, not blocking.

Piyush-85 (commented):
Recommend adding token revocation on logout (`POST mtls_endpoint_aliases.revocation_endpoint`). Without it, refresh tokens remain valid at the authorization server after a user logs out, which is a security concern for certificate-bound sessions. This can land in this PR or a tracked follow-up — if it's deferred, please link the follow-up.

lib/config.js
```diff
@@ -373,15 +386,65 @@ module.exports.get = function (config = {}) {
...
+          'global has no client certificate API.',
+      );
+    }
+
+    // mTLS replaces secret/assertion-based client auth. Allowing both is a
+    // misconfiguration that would not use the certificate as intended.
+    if (value.clientSecret || value.clientAssertionSigningKey) {
+      throw new MtlsError(
+        MtlsErrorCode.MTLS_INCOMPATIBLE_CLIENT_AUTH,
+        'useMtls cannot be combined with clientSecret or clientAssertionSigningKey. ' +
```
Piyush-85, line 446: Nit: `clientSecret` is always seeded from `process.env.CLIENT_SECRET` (line 389), so enabling `useMtls` with a stale `CLIENT_SECRET` in the environment throws `MTLS_INCOMPATIBLE_CLIENT_AUTH`. Correct behavior, but the message doesn't hint the value may have come from the env. Consider mentioning env vars to speed up debugging.

lib/context.js
jd3vi1: Fixed in e6cac71. Extracted the check into a `warnIfNotCertificateBound(config, accessToken)` helper and now call it from both the callback path and `refresh()` (after the new `access_token` is written into the session), so a token that loses its `cnf.x5t#S256` binding on refresh now warns too.

lib/context.js
jd3vi1: Good catch on the mismatch with the Testing Doc. Rather than add a JWE-specific warning, we've dropped the JWE handling from scope here: JWE access-token decryption is a Resource Server concern (tracked separately, and PR #864 was closed), so this SDK does not inspect encrypted access tokens. Updated the comment wording accordingly in e6cac71 (now reads "Opaque (non-JWT) access tokens ... are skipped", no JWE mention).

lib/config.js
jd3vi1: Done in e6cac71. The `MTLS_INCOMPATIBLE_CLIENT_AUTH` message now notes that `clientSecret` may have been seeded from the `CLIENT_SECRET` environment variable and suggests unsetting it.

jd3vi1:
@Piyush-85 on token revocation on logout: agreed it's a valid concern, but we're deferring it to a separate PR rather than including it here.

Two reasons: (1) it isn't mTLS-specific. This repo has never revoked the refresh token on logout for any client-auth method; `logout()` clears the local session and optionally does federated logout (`idpLogout`), which ends the session server-side at the AS. So this is a pre-existing, SDK-wide behavior, not something mTLS introduces. (2) Adding RFC 7009 revocation is a standalone feature (new config surface, endpoint routing including the mTLS alias, error type, tests) that doesn't belong in a client-authentication PR.

It doesn't exist anywhere in this repo today, so we'll track it as its own follow-up rather than link an existing one. The other three comments (refresh-path binding check, JWE wording, env-var hint in the error message) are addressed in e6cac71.

Piyush-85 (dismissed):

Piyush-85 (approved):
LGTM ! Reviewed it async with @jd3vi1

PR #873, opened 2026-08-07
feat: Anonymous Sessions support

cschetan77:
## Description

Adds opt-in support for **Anonymous Sessions**, which let an application give a visitor an Auth0 identity before they log in. A visitor gets an anonymous session (backed by an `auth0_anon` cookie) and can obtain access tokens to call APIs; when they later authenticate, the anonymous session token is automatically carried into the authorization request so it can be linked to the authenticated identity.

The feature is disabled by default and fully additive — existing behavior is unchanged when `anonymousSession.enabled` is not set.

### Configuration

```js
app.use(
  auth({
    authRequired: false,
    anonymousSession: {
      enabled: true,
      // cookie options are optional and mirror the session cookie defaults
      // (name defaults to `auth0_anon`)
    },
  }),
);
```

### Request context

When enabled, every request gets a `req.anonymousSession`:

- `req.anonymousSession.isAnonymous` — `boolean`
- `req.anonymousSession.token` — the opaque anonymous session token, or `null`
- `await req.anonymousSession.start({ audience?, scope?, metadata? })` — creates a session and returns its access token (metadata is set once, at creation)
- `await req.anonymousSession.getAccessToken()` — returns the cached access token while valid, silently renewing it (or transparently creating a new session if the anonymous session has expired) when it is not
- `await req.anonymousSession.end()` — ends the session and clears the cookie

`AnonymousSessionError` (exported from the package) is thrown for surfaced failures and carries a `.code`.

## How it works

- **HTTP client** (`lib/anonymousSession/client.js`) calls `POST /anonymous/token` and `POST /anonymous/logout`. It reuses the existing custom-fetch wrapper so the telemetry and `User-Agent` headers (and the `enableTelemetry` opt-out) are preserved, and derives the endpoint from the discovered issuer. Client authentication is sent based on the configured `clientAuthMethod` (Basic for confidential clients, `client_id` in the body for public clients).
- **Cookie** (`lib/anonymousSession/cookie.js`) stores the anonymous session state in a single encrypted `auth0_anon` cookie, using the same encryption profile as the application session cookie. Only the session token is ever forwarded to the authorization server; the access token stays server-side. Renewals do not extend the cookie lifetime.
- **Context** (`lib/anonymousSession/context.js`) implements the read accessors and the `start`/`getAccessToken`/`end` methods. Cookie writes are deferred to an `on-headers` hook so request-side methods never touch the response directly, matching the application-session pattern.
- **Login** injects the anonymous session token into the `/authorize` request from the existing login handler; the developer calls login as usual.

Anonymous state is kept completely independent of `req.oidc.isAuthenticated()` and `req.oidc.user`.

## Testing

- New unit tests for config, the HTTP client, the cookie handler, and the request context, plus end-to-end integration tests through an Express server (all HTTP stubbed with `nock`).
- `index.d.ts` types and `tsd` type tests updated.
- `npm run test` and `npm run lint` pass.

## Checklist

- [x] Additive, opt-in, no breaking changes
- [x] Types kept in sync (`index.d.ts`)
- [x] Unit + integration tests added
- [x] `EXAMPLES.md` and a runnable example updated

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -16,6 +16,7 @@
 14. [Use a proxy for OIDC requests](#14-use-a-proxy-for-oidc-requests)
 15. [Session expiry from upstream IdP (IPSIE `session_expiry`)](#15-session-expiry-from-upstream-idp-ipsie-session_expiry)
 16. [JWT-Secured Authorization Requests (JAR)](#16-jwt-secured-authorization-requests-jar)
+17. [Anonymous Sessions](#17-anonymous-sessions)
 
 ## 1. Basic setup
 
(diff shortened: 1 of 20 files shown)


cschetan77:
Closing this PR as we’re not planning to implement Auth0-specific features in this SDK.
If you want Auth0-specific features, we recommend trying out the [@auth0/auth0-express](https://github.com/auth0/auth0-express) SDK, currently in beta.

PR #880, opened 2026-08-24
feat(enterprise-connect): app-embedded Enterprise Connect support

jd3vi1:
## Summary

Adds SDK support for Enterprise Connect's app-embedded integration pattern ([SDK-10114](https://auth0team.atlassian.net/browse/SDK-10114)), where Auth0 acts as a pure SSO relay — it federates to the customer's enterprise IdP and hands back an ID token, without creating or holding an Auth0 session. The application owns the session entirely.

Spec: [\[SDK-10114\] Enterprise Connect - Embedded Login (express-openid-connect)](https://oktainc.atlassian.net/wiki/spaces/~712020a022b5e295fc4b2d9b845fb141ea43e3/pages/1057063684)

## What's included

- **`isFederatedDomain(auth0Domain, emailDomain)`** — standalone exported async helper. WebFinger-based domain discovery (`urn:auth0:discovery:domain:{domain}` resource), with response caching: `true` for 60s, `404`-derived `false` for 15s, FIFO-capped at 1000 entries. Ambiguous/403/429/5xx/network-error responses are `false` and never cached; 429 also emits `console.warn`.
- **`enterpriseConnect: true`** config option. Puts the SDK into Enterprise Connect mode. Warns at init if `offline_access` is in scope (no refresh tokens are issued) or if a static `organization` is configured (it's resolved per login via Home Realm Discovery from `login_hint` instead).
- **`afterCallback` stateless passthrough** — returning `null`/`undefined` now suppresses the SDK's own session write. This also **fixes a pre-existing crash**: returning `null`/`undefined` while a different user was replacing an existing session threw a `TypeError` in `replaceSession()`, since it sets a property directly on its argument.
- **`startEnterpriseLogin({ email, returnTo })`** — new method on `res.oidc`, alongside `login`/`logout`. Single login entry point composing domain discovery with a `login_hint`-based redirect. No `connection`/`organization` lookup required — Auth0's HRD resolves both from the email domain.
- **EC-aware logout** — when `enterpriseConnect: true`, `res.oidc.logout()` bypasses the existing `auth0Logout`/v2-vs-OIDC selection entirely, builds the `end_session_endpoint` URL directly (falling back to `/oidc/logout` if none is advertised), forces `federated=true` as a literal string when requested, omits `id_token_hint`/`logout_hint` (no Auth0 session exists to source them from), and clears any dangling transaction cookie.

## Usage

```js
const { auth, isFederatedDomain } = require('express-openid-connect');

app.use(
  auth({
[...]

PR #18, opened 2019-09-25
Fix sec warnings for dependencies

hzalaz:
Updated the following libraries

* @sinonjs/text-encoding
* argparse
* eslint-utils
* esprima
* js-yaml
* just-extend
* lodash
* lodash.merge
* @sinonjs/samsam
* @sinonjs/formatio
* lolex
* nise

PR #881, opened 2026-08-24
feat(enterprise-connect): app-embedded Enterprise Connect support

jd3vi1:
## Summary

Adds Enterprise Connect app-embedded support. Auth0 acts as a pure SSO relay: it federates to the enterprise IdP and hands back an ID token, but creates no session. The application owns the session entirely.

## What's included

- `isFederatedDomain(auth0Domain, emailDomain)`: WebFinger domain discovery. Cached (true for 60s, 404-derived false for 15s, FIFO cap at 1000). Ambiguous/403/429/5xx/network-error results are false and never cached.
- `enterpriseConnect: true` config option. Warns at init on `offline_access` in scope, and on a static `organization` (it's resolved per login via Home Realm Discovery instead).
- `afterCallback` can return `null`/`undefined` to skip the SDK's own session write. This also fixes a real crash: doing so while a different user replaces an existing session used to throw in `replaceSession()`.
- `startEnterpriseLogin({ email, returnTo })` on `res.oidc`. Single login entry point. No `connection`/`organization` lookup needed; HRD resolves both from the email domain.
- EC-aware `res.oidc.logout()`: uses the standard `end_session_endpoint` directly (skips the existing `auth0Logout`/v2 selection), sends `federated=true` as a literal string when requested, sends no `id_token_hint`/`logout_hint`.

## Usage

```js
const { auth, isFederatedDomain } = require('express-openid-connect');

app.use(
  auth({
    issuerBaseURL: `https://${process.env.AUTH0_DOMAIN}`,
    clientSecret: process.env.CLIENT_SECRET,
    secret: process.env.SECRET,
    enterpriseConnect: true,
    authorizationParams: {
      response_type: 'code',
      scope: 'openid profile email',
    },
    afterCallback: async (req, res, session) => {
      const claims = req.oidc.idTokenClaims;
      await mySessionStore.upsert({ sub: claims.sub, email: claims.email, orgId: claims.org_id });
      return undefined;
    },
  }),
);

app.post('/login', express.json(), async (req, res) => {
  const started = await res.oidc.startEnterpriseLogin({ email: req.body.email, returnTo: '/dashboard' });
  if (!started) res.redirect('/existing-login');
});

app.get('/logout', (req, res) => {
  mySessionStore.destroy(req);
  res.oidc.logout({ returnTo: process.env.BASE_URL, federated: true });
});
```

## Tests

473 passing (18 new/extended). `tsd`, `eslint`, `prettier` all clean.

- New: `test/isFederatedDomain.tests.js`, `test/startEnterpriseLogin.tests.js`
- Extended: `test/callback.tests.js`, `test/config.tests.js`, `test/logout.tests.js`, `index.test-d.ts`
- Docs: `EXAMPLES.md` #18, runnable `examples/enterprise-connect.js`

## Notes for reviewers

- `login`, `logout`, and `startEnterpriseLogin` are all called on `res.oidc`, not `req.oidc`. The design doc's Express snippets use `req.oidc`, which is wrong for this SDK's actual convention.
- `isFederatedDomain`'s cache avoids private class fields and numeric separators, since this repo's eslint config targets ES2020.
- WebFinger is blocked until `flags.local_resource_discovery` is enabled tenant-side. All tests run against `nock`; end-to-end verification against a live tenant is blocked on that.

Supersedes #880, which GitHub auto-closed when its source branch was renamed.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -852,7 +852,7 @@ app.get('/dashboard', async (req, res) => {
 
 ### Logout
 
-Destroy your own session, then delegate to the SDK's logout with `federated: true` so the enterprise identity provider session is terminated as well. Without it, the IdP session stays alive and the next login silently reuses the previous user. The SDK builds the correct `end_session_endpoint` URL directly in this mode - there is no Auth0 session to source an `id_token_hint` from, so none is sent.
+Destroy your own session, then delegate to the SDK's logout with `federated: true` so the enterprise identity provider session is terminated as well. Without it, the IdP session stays alive and the next login silently reuses the previous user. There is no dedicated Enterprise Connect logout path: since the SDK never wrote its own session in this mode, `id_token_hint` is naturally absent, and `federated: true` flows through the same logout mechanism used everywhere else.
 
 ```js
 app.get('/logout', async (req, res) => {
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -17,6 +17,7 @@
 15. [Session expiry from upstream IdP (IPSIE `session_expiry`)](#15-session-expiry-from-upstream-idp-ipsie-session_expiry)
 16. [JWT-Secured Authorization Requests (JAR)](#16-jwt-secured-authorization-requests-jar)
 17. [mTLS client authentication](#17-mtls-client-authentication)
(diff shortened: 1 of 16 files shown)


jd3vi1:
Closing this since it's an auth0 specific feature. 
This will be built in auth0-express

